PR #55, opened 2015-01-05
Refactoring $RTP_IP/$RTP_IP6 so we can use a number of parameters for --interface command line

sybasesql:


rfuchs:
Thanks for this. Do you think it may be better to get rid of $RTP_IP6 altogether, since it does the same thing as $RTP_IP?

rfuchs:
Actually, nevermind my last comment. The init script should still support it for compatibility nevertheless.

sybasesql:
Yep! It'll even better and simpler.

sybasesql:
Well, with this pull request there are changes in the structure of the configuration file (rtpengine.sysconfig), so changes will be anyway and i afraid not backward compatibility.
e.g. deleting $RTP_ADV_IP/$RTP_ADV_IP6 and transforming $RTP_IP/$RTP_IP6 into array.

rfuchs:
Ah yes, you're right. I'll go ahead and remove it then.

PR #69, opened 2015-02-03
Some more features from 1and1

inf265:


rfuchs:
Hi,

First of all, thank you. As usual, a few comments from my side. :)

There's locks missing in a few places. There's the total_stats, as already discussed. Some of these operations could be replaced through atomic ops (g_atomic_int_add etc), but others can't and concurrency will mess up the stats sooner or later.

The new control_ng_stats also needs either locking or atomic ops, both for maintaining/managing the list itself, and for the contained stats. I would also highly suggest using a hash table instead of a linked list for control_ng_stats (which would definitely require a mutex).

The graphite code can leak file descriptors in certain errors cases (sockets opened but never closed).

The graphite loop function needs a shorter sleep time. The problem is during shutdown: threads aren't cancelled, but are waited for to terminate. If the thread sleeps for 60 seconds, then it can take that long for the daemon to shut down. I would suggest sleeping less than one second at a time (100 ms works fine), and using time() or similar to control the interval (see poller_timers_wait_run() for example). Alternatively you can use the timer mechanisms in poller.c instead of a separate thread (like callmaster_timer() does), but keep in mind that there's only one timer thread running, so for any blocking operation you should launch a one-shot thread.

Thanks!

inf265:
Danke für den Kommentar. Ich schau mir das mal an. Evtl. hätte ich noch ein paar Rückfragen, insbesondere bzgl. der Nutzung der hashmap, da benutzt du glaub ich die Glib, richtig ? Das muss ich mir erstmal ansehen.

rfuchs:
The required changes aren't very big, I can also do them myself if it's too much trouble for you.

inf265:
Hi Richard,
that would be _great_ from you ! So I can see how you usually do it and for the next features I can adopt it. I am not very used with the glib locking stuff, only worked with mutex's from the posix stuff.
Thanks a lot.
Cheers, Frédéric

rfuchs:
The locking is really the same as POSIX (it uses pthread_* stuff).

Commits are coming in here and will be merged to master when done:
https://github.com/sipwise/rtpengine/tree/1and1-master

PR #101, opened 2015-04-22
Latest updates from 1&1

inf265:
Including some other features and fixes.


rfuchs:
Just to make sure I understand the "measuredelay" mechanism right: It's supposed to measure the delay between receiving the packet and sending it back out? And in kernel space only?

rfuchs:
I've been looking at the kernel module code and it seems that it can be much simplified by making use of sk_buff->tstamp, which (as far as I can tell) is automatically set by the kernel when the packet is received, or otherwise it can be set explicitly. It also has nanosecond precision and the delay can be calculated by subtracting from ktime_get_real(). Would that work for your purposes?

inf265:
Wow great. I hadn't in mind that the kernel also sets the time in the
sk_buf. You suggestion is better. That'll definitely work for our
purpose.Thanks a lot.
- Frédéric

Am 06.05.15 um 17:21 schrieb Richard Fuchs:

Issue #102, opened 2015-04-22
Aastra + SRTP/RTP no audio after hold


GGGO:
Hi Richard,

we already talked about it, and I would like to create an issue for it.

Aastra ---SRTP---> rtpengine ---RTP---> rtpproxy---RTP--->asterisk

Note: rtpproxy will be removed soon, but for now I can't.

The problem is the following:
A calls B(Aastra), B(Aastra) answers A, 
B(Aastra) receives another call from C, B(Aastra) puts on hold A and answers C
B(Aastra) put on hold C and comes back to A
And now B(Aastra) can't hear A, but A can hear B(Aastra)

When I listen the call recording in asterisk, I can hear A and B(Aastra), so audio arrive without any problem.

It's difficult to reproduce the problem, I could make 10 or more tests without seing it.
It's happens sometimes for whatever reason.

But I've found this https://issues.asterisk.org/jira/browse/ASTERISK-20194, and it appears to be the same problem I have. Matt Jordan resolved it by applying 2 patches. See srtp_patches.diff

It seems Aastra change its crypto key, and I think rtpengine doesn't reset the crypto key if the endpoint hasn't change. So rtpengine will use on old crypto key, and Aastra will never success to decrypt the audio.

static void __fill_stream(struct packet_stream *ps, const struct endpoint *epp, unsigned int port_off) {
	struct endpoint ep;

	ep = *epp;
	ep.port += port_off;

	/* if the endpoint hasn't changed, we do nothing */
	if (PS_ISSET(ps, FILLED) && !memcmp(&ps->advertised_endpoint, &ep, sizeof(ep)))
		return;

	ps->endpoint = ep;
	ps->advertised_endpoint = ep;
	/* we reset crypto params whenever the endpoint changes */
	crypto_reset(&ps->crypto);
	dtls_shutdown(ps);
	PS_SET(ps, FILLED);
}

Maybe we need to keep the remote crypto key somewhere in the rtpengine session, and if the crypto has changed, we reset it ?

GGGO:
I just commented this 2 lines to force crypto_reset() every times:

/*if (PS_ISSET(ps, FILLED) && !memcmp(&ps->advertised_endpoint, &ep, sizeof(ep)))
    return;*/

Will see if it's better.

thomaswardiii:
The Asterisk bug appears to be the opposite of what you are experiencing.

[...]

PR #107, opened 2015-04-29
reset SRTP ROC to zero when new crypto incoming

GGGO:
workaround to https://github.com/sipwise/rtpengine/issues/102, for broken audio after SSRC changed by
hold/unhold

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -1924,8 +1924,11 @@ static int __init_stream(struct packet_stream *ps) {
 	int active;
 
 	if (ps->sfd) {
-		if (MEDIA_ISSET(media, SDES))
+		if (MEDIA_ISSET(media, SDES)) {
+			if (crypto_params_cmp(&ps->sfd->crypto.params, &media->sdes_in.params))
+				ps->crypto.last_index = 0;
 			crypto_init(&ps->sfd->crypto, &media->sdes_in.params);
+		}
 
 		if (MEDIA_ISSET(media, DTLS) && !PS_ISSET(ps, FALLBACK_RTCP)) {
 			active = (PS_ISSET(ps, FILLED) && MEDIA_ISSET(media, SETUP_ACTIVE));


rfuchs:
I've considered a patch like this, but it would be incorrect. RFC 3711 specifically states in 3.3.1:

```
   After a re-keying occurs (changing to a new master key), the rollover
   counter always maintains its sequence of values, i.e., it MUST NOT be
   reset to zero.
```

GGGO:
That's true only when the same SSRC is used, no ?

Aastra always changes its SSRC. Yealink uses the same when doing hold/unhold. 

We should check if SSRC has changed before resetting the ROC, but right now I can't get the rtp header in __init_stream and compare the SSRC value. 

So this patch works with Aastra but will breaks phones which use the same SSRC when re-keying.

rfuchs:
SSRC and destination address are the only criteria. Difference in master key is irrelevant.

SSRC change should be addressed with https://github.com/sipwise/rtpengine/commit/39a305f4badd3bc43d2d5239e0112006ce2540d2

GGGO:
Yes exactly, but it's my only way to detect a SSRC changed right now. 

I would use rtp_savp2avp to check the SSRC but I won't be able to reset the last_index of the other leg context because in this function we only have stream->sfd->crypto, I need the stream->crypto context.

rfuchs:
I understand that it fixes it for you, but it's still an incorrect patch.

rtp_savp2avp already includes logic to guess the ROC if packets can't be decrypted (as determined by a failed authentication hash) and so shouldn't need to track the SSRC.

rtp_savp2avp also shouldn't need to touch the ROC of the other leg if the SSRC changes since they're two distinct crypto contexts with two different SSRC and ROCs. If your phone resets the ROC in both directions if one SSRC changes, then that's a bug in your phone.

GGGO:
Yes you'r right.

rfuchs:
If you're unable to find a different (proper) solution to this, then I'm willing to accept a patch like this as an optional feature, either selectable on a per-call basis or as a global switch or compile time option.

GGGO:
Ok that will be great. I will send a bug report to Aastra too.

Thank you.

PR #110, opened 2015-05-08
Resolves ROC inconsistancies when encrypting SRTP

marcantonio:
@mclazarus and I resolved https://github.com/sipwise/rtpengine/issues/102.  We addressed the following:

1. Uses the proper ROC in kernel space when encrypting.
2. Track all SSRCs we've seen along with the last ROC associated with each.  If the SSRC is used again we revert the ROC, otherwise it's zeroed out.
3. Re-implemented RFC 3711 index determination to avoid a few dead spots in the sequence numbers.

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -161,6 +161,7 @@ static int call_avpf2avp_rtcp(str *s, struct packet_stream *);
 static int call_savpf2avp_rtcp(str *s, struct packet_stream *);
 //static int call_savpf2savp_rtcp(str *s, struct packet_stream *);
 
+static void unkernelize(struct packet_stream *);
 
 /* ********** */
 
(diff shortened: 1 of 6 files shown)


rfuchs:
So I've changed a few things around, trying to make things easier for the most common case (SSRC doesn't change). Everything is committed here: https://github.com/sipwise/rtpengine/tree/marcantonio-ssrc-seq

Perhaps you can run this against your test case to see if everything still works for you.

marcantonio:
I ran it through most of our tests and it looks good.  I left one question for you on your changes.  The rest makes sense.  Thanks!

mclazarus:
Thanks for your help on this @rfuchs !

PR #136, opened 2015-07-16
Bypass the randomization of get_consecutive_ports()

smititelu:
One solution(the one presented) is to add two more wanted_ports to the monologue_offer answer() will eventually call get_consecutive ports for each of the wanted port.
 
Another solution to this bypassing would be to retain the wanted_ports in a structure (like callmaster) and call the get_endpoint_map()->get_consecutive_ports() with the values found there(defaulted to 0).

What do you think about it?

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -1750,7 +1750,7 @@ struct stream_fd *__stream_fd_new(struct udp_fd *fd, struct call *call) {
 }
 
 static struct endpoint_map *__get_endpoint_map(struct call_media *media, unsigned int num_ports,
-		const struct endpoint *ep)
+		const struct endpoint *ep, unsigned int wanted_start_port)
 {
 	GSList *l;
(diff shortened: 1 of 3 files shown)


rfuchs:
I don't think this is going to work reliably. In cases where a dialogue contains multiple media streams (e.g. audio and video), you have no way of determining/controlling the ports of the second (and later) streams.

pkuzak:
Indeed you are right. I have to admit that we just didn't think about having more than one stream as we mostly only deal with audio. We have to think of a better solution here.
Nonetheless, assuming one only has to deal with audio and there are never other streams, would you consider the proposed changes sufficient in order to get the wanted bridge ports assigned by _rtpengine_ for a specific stream?

rfuchs:
I suppose so.

pkuzak:
Thank you for your input here. From my side you can close this pull request.

PR #139, opened 2015-07-27
Added a bridgeport GQueue to the 'struct call'

smititelu:
If the queue has elements, then those will be considered as bridgeports.

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -1750,7 +1750,7 @@ struct stream_fd *__stream_fd_new(struct udp_fd *fd, struct call *call) {
 }
 
 static struct endpoint_map *__get_endpoint_map(struct call_media *media, unsigned int num_ports,
-		const struct endpoint *ep)
+		const struct endpoint *ep, unsigned int wanted_start_port)
 {
 	GSList *l;
 	struct endpoint_map *em;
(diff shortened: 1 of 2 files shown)


daemon/call.c
```diff
@@ -2466,8 +2466,10 @@ int monologue_offer_answer(struct call_monologue *other_ml, GQueue *streams,
 	struct call_monologue *monologue = other_ml->active_dialogue;
 	struct endpoint_map *em;
 	struct call *call;
+	unsigned int bridge_count;
```
rfuchs: You can eliminate this, as g_queue_pop_head returns NULL if the queue is empty anyway, which translates to zero

daemon/call.c
```diff
@@ -2595,7 +2597,12 @@ int monologue_offer_answer(struct call_monologue *other_ml, GQueue *streams,
 
 		/* get that many ports for each side, and one packet stream for each port, then
 		 * assign the ports to the streams */
-		em = __get_endpoint_map(media, num_ports, &sp->rtp_endpoint);
+		if (!bridge_count) {
+			em = __get_endpoint_map(media, num_ports, &sp->rtp_endpoint, 0);
+		} else {
+			unsigned int *bridge_port = (unsigned int *)g_queue_pop_head(&call->rtp_bridge_ports);
```
rfuchs: These are supposed to be dynamically allocated "int"s? You can skip this detour by casting the int directly to a pointer and then back, which also eliminates possible memory leaks. See GINT_TO_POINTER() and GPOINTER_TO_INT() or just typecast it yourself.

rfuchs:
See my inline comments for simplifying this. Otherwise OK

daemon/call.c
smititelu: They were static uints from a dinamically allocated array of uints. I thought that q_queue_push_tail() needed actually the uint's address but I see now that I could have called it with the actual value (casted to gpointer). Thanks for the suggestion. Also looks clearer using glib macros.

daemon/call.c
smititelu: You are right. I've eliminated it.

daemon/call.c
smititelu: Simplified it also taken into consideration that g_queue_pop_head() returns NULL (0) if the queue has no elements.

PR #146, opened 2015-08-23
rtpengine.init now working with firewalld

divereigh:
Now works with firewalld - creates a permanent chain called INPUT_prefilter
which is then called before the conntrack stuff in the INPUT tables. We then create
our temporary rtpengine chain by linking into that.
Also brought in some of the IPv6 stuff from the standard Centos init script

### DIFF
--- a/el/rtpengine.init
+++ b/el/rtpengine.init
@@ -56,10 +56,13 @@ build_opts() {
 	fi
 	shopt -u nocasematch
 
-	if [[ -n "$RTP_IP" ]]
-	then
-		OPTS+=" --interface=$RTP_IP"
-	fi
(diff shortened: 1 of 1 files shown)


rfuchs:
Thanks for this. I've noticed you've brought back usage of IPv6-specific options (`$RTP_IP6` etc). These have been obsoleted in the sysconfig file a while ago. If you want to avoid creating rules for IPv6 when it's not in use, then this needs to be detected in some other way.

divereigh:
Yup no worries, I have removed them. I started from the centos 7 version of the init script, so I wasn't sure if those options had been added - should have just stuck to the bits I knew about :-)

PR #147, opened 2015-08-24
Rtpengine bind on specific IP address

smititelu:
Right now rtpengine binds on any IP address. Add posibility to bind only on the
IPs given by -i parameter. Extended this for both IPv4 and IPv6.

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -1555,23 +1555,30 @@ static void __get_pktinfo(int fd) {
 	setsockopt(fd, IPPROTO_IP, IP_PKTINFO, &x, sizeof(x));
 }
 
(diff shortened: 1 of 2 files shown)


smititelu:
Tested for single IPv4 --interface. Tested on a  --interface=pub/192.0.2.67 --interface=priv/10.35.2.75 audio scenario which worked; 2 bridge ports were opened on each interface.

rfuchs:
Thanks, but the changes stashed away in `rfuchs/socket-rework` (which will be merged in eventually) already implement this.

This patch doesn't support interface switching (as required by ICE for example) so this is a no-go.

PR #158, opened 2015-09-23
Specific REDIS read and write databases

smititelu:
The REDIS_READ and REDIS_WRITE params have priority over REDIS. REDIS can substitute any or both of them.

Added a new "role" parameter which can help to force connecting only to a specific master/slave redis. If the connected redis role doesn't match the wanted role, the connection is retried and eventually fails. Using "ANY_REDIS_ROLE" one does not care about the role of the connected redis. This is needed for maintaining the current logic for REDIS, which connects only to master redises, while making possible a master-slave redis setup using REDIS_READ+REDIS_WRITE.


smititelu:
Solved conflict after 6db99f9. Able to merge.

rfuchs:
If I read this correctly, this is an exclusive either/or scenario, meaning either a single redis DB (-z/-Z) is specified, or one read DB (-r/-R) and one write DB (-w/-W), but never both, correct? If so, any objections against retaining -r/-R for single redis DB mode, and use the same parameter for the read redis DB if -w/-W are also specified? This would make things a bit simpler and also keep the CLI compatible with old versions.

smititelu:
One can specify combinations of both REDIS + REDIS_READ/WRITE at the same time. Also REDIS + REDIS_READ + REDIS_WRITE at the same time. Also just REDIS_READ or just REDIS_WRITE. If a more specific REDIS_READ/WRITE is present, it overwrites the REDIS for that particular action(read/write).

Does your proposal still applies to this behaviour? Do you have other behaviour proposals?

smititelu:
Updated pull request according to the second suggestion. CLI backwards compatibility is maintained by using -r for REDIS and -z for REDIS_READ.

PR #163, opened 2015-10-02
Bug fix when used fd is 0

smititelu:
Default fd to -1 after alloc a new struct socket_t in order to avoid closing fd 0. Add some logs.


rfuchs:
Good catch, thanks.

PR #165, opened 2015-10-07
Return error when no free ports are available

smititelu:
Free the used structures and sockets and  return -1 if there are no free ports on _all_ local interfaces of the given logical interface.

Suppose the case of audio+video; suppose 4 ports are available, so bridgeports are allocated for the audio stream; now suppose next 4 ports are not available anymore(for video). In this case of free ports failure one needs also to free the audio sockets also. In order to do this, I called call_destroy() upon get_consecutive_ports() failure.

The other approach would have been to implement functions that iterate through monologue's medias and free the sockets, but I've found call_destroy() more clean.

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -1595,7 +1595,8 @@ init:
 	return 0;
 
 error:
-	ilog(LOG_ERR, "Error allocating media ports");
+	ilog(LOG_ERR, "Error allocating media ports, destroying call %.*s", call->callid.len, call->callid.s);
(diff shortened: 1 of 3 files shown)


rfuchs:
If you have no objections, I will apply this with some of the logging changes reverted. ilog() already included the call-id in the prefix for each log message, so including the call-id again is kinda redundant.

smititelu:
You are right. No objections from my side.

PR #166, opened 2015-10-07
Round Robin logical interface selection based on direction attribute

smititelu:
If **direction=round-robin-calls** attribute is received, a round robin algorithm runs for choosing the logical interface of the call media. The algorithm checks that _all local interfaces_ of the tried logical interface have PORTS_PER_STREAM=4 free ports. If a logical interface fails the check, the next one is tried. If there is _no_ logical interface found with this property, it fallbacks to the default behaviour (e.g. return NULL and on the subsequent call and eventually first logical interface in INTERFACES is returned even if no ports are available). The attribute is ignored for answers() because the logical interface was already selected at offers().

In the current implementation, in the case of audio+video streams, the audio stream(4 ports) might be assigned on log1 and the video stream(4 ports) on log2. This is because I tried to keep the changes as simple as I could. If one wants forcing audio + video on the same log1, the header of the functions must be modified to also accept the number of ports to be checked as free(e.g. 8 ports) and decreasing them while the streams get allocated(e.g. check for 8 ports for audio stream, then check for another remaining 4 ports for the video stream).

The purpose of the round robin is to load balance between a list of logical interfaces given.

### DIFF
--- a/README.md
+++ b/README.md
@@ -710,6 +710,25 @@ Optionally included keys are:
 	However, this mechanism for selecting the address family is now obsolete
 	and the `address family` dictionary key should be used instead.
 
+	A direction keyword is *round-robin-streams*. If this is received, a round robin algorithm runs for
+	choosing the logical interface for the current stream(e.g. audio, video).
+	The algorithm checks that all local interfaces of the tried logical interface have free ports for a
+	stream (e.g. 4). If a logical interface fails the check, the next one is tried. If there is no logical
+	interface found with this property, it fallbacks to the default behaviour (e.g. return first logical
+	interface in --interface list even if no free ports are available). The attribute is ignored for
+	answers() because the logical interface was already selected at offers().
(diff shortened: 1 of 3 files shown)


smititelu:
Solved conflicts. Updated README with some examples. Changed the keyword name to *round-robin-streams* as it's more like the case.

smititelu:
Created a wikipage with some known limitations on the current implementation of round robin: https://github.com/sipwise/rtpengine/wiki/Round-Robin

smititelu:
Closing this due to some bugs found when testing.

rfuchs:
I would prefer it if the number of ports wasn't hardcoded, but rather dependent on how many ports are actually needed. Also, iterating through the entire port range on every request is not a good approach, as it would start wasting CPU time once the port ranges become congested. It would be better to just keep a counter that gets updated as ports become allocated and deallocated.

PR #171, opened 2015-10-16
Allow more than 65536 open files

smititelu:
Make the maximum number of open files (in /proc/pid/limits) configurable using "ulimit -n X" command.
Needed for scenarios in which rtpengine binds to more than one IP.


smititelu:
Updated pull request. Initial open-files upper limit changed to 1 << 20.
Made this number configurable using "rtpengine-ctl set max-open-files <X>". The minimum configurable through ctl command is 1 << 16.

PR #175, opened 2015-10-27
Round Robin logical interface selection

smititelu:
Improvements since last time:
- maintain and update volatile _free_ports_ while binding and releasing ports => fast deciding if free ports available
- no hard coded #defines; the ports are calculated based on the stream_params that have "round-robin-calls" direction
- take into consideration the media call legs based on given kamailio direction (e.g. direction=round-robin-calls direction=pub1)

### DIFF
--- a/daemon/media_socket.c
+++ b/daemon/media_socket.c
@@ -28,8 +28,6 @@
 #endif
 
 
-
-
 typedef int (*rewrite_func)(str *, struct packet_stream *);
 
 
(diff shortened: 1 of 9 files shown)


daemon/media_socket.c
```diff
@@ -418,13 +570,24 @@ static int get_port(socket_t *r, unsigned int port, struct intf_spec *spec) {
 		return ret;
 	}
 
+	pp->free_ports--;
```
rfuchs: I think all operations on the free_ports var need to use atomic ops - volatile alone isn't enough. The g_atomic_int_* functions should be used, at least for inc and dec operations (but better to use them for everything)

daemon/media_socket.c
```diff
@@ -226,14 +224,140 @@ static GHashTable *__logical_intf_name_family_hash;
 static GHashTable *__intf_spec_addr_type_hash;
 static GQueue __preferred_lists_for_family[__SF_LAST];
 
+static unsigned int selection_index = 0;
```
rfuchs: These 2 vars are also global and so need some synchronization. It looks like converting them to thread-local storage (using __thread) would be enough.

daemon/media_socket.c
```diff
@@ -226,14 +224,140 @@ static GHashTable *__logical_intf_name_family_hash;
...
+	GList *log_iter;
+
+
+	nr_logs = g_queue_get_length(q);
+	log_iter = q->head;
 
+select_log:
+	// choose the next logical interface
+	index = 0;
+	while (index < selection_index && log_iter) {
```
rfuchs: This can be shortened a bit by using g_queue_peek_nth() or g_queue_peek_nth_link()

rfuchs:
A few minor comments posted - overall looks good to me.

smititelu:
Updated pull request for the above comments.

You are right, using __thread is the best thing we could do here because we don't want any other synchronization on those globals, which might add lag to rtpengine (that's why I didn't added any sync in the first place and injection tests were ok). However, because the variables become per thread, bindings to the same IP for 2 consecutive calls _might_ happen.

Basically tested it and it was ok (I could also see the above behaviour). Before merging this, please wait for @pkuzak feedback as he is doing some tests closer to real life scenarios (also pushing this feature to the limit).

pkuzak:
From QA perspective I have no objections. Successfully tested this feature with 20k calls using 2 IPs.

rfuchs:
I've added a few more commits after the merge, make sure you do a pull/rebase

PR #176, opened 2015-10-28
Avoid cdr buffer overlows. Add some ICE NULL checks

smititelu:
sprintf might lead to overflows.
Encountered and ice_shutdown() segfault when call was destroyed (suppose cdr buffer overflow caused it, corrupting pointer values).


rfuchs:
You may consider a dynamically allocated and automatically growing buffer for the cdr logs for the future, e.g. asprintf, a GString or something like that.

PR #178, opened 2015-10-29
Allow non-root users to run rtpengine in kernel mode

camilleoudot:
This patch adds parameters to the kernel module, to choose the userid and groupid of the files in the /proc/rtpengine tree, allowing non-root users to run and control rtpengine.

### DIFF
--- a/el/rtpengine.init
+++ b/el/rtpengine.init
@@ -180,7 +180,12 @@ start() {
 	then
 		echo "Loading module for in-kernel packet forwarding"
 		rmmod xt_RTPENGINE 2> /dev/null
-		modprobe xt_RTPENGINE
+		if [[ -n "$RE_USER" ]]
+		then
+			modprobe xt_RTPENGINE set_proc_id=1 proc_uid=$(id $RE_USER -u) proc_gid=$(id $RE_USER -g)
+		else
+			modprobe xt_RTPENGINE
+		fi
 		temp=`firewall-cmd --state 2>/dev/null`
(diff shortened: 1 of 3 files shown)


rfuchs:
Very nice, thank you.

Two questions here, does it make sense to have a master on/off switch for this feature, when the default uid/gid is 0/0 anyway? And, while we're at it, how about adding an umask for all pseudofiles, so that e.g. access could be controlled through the group?

camilleoudot:
I do agree about the master switch, I planned to remove it eventually, can do it now.
Regarding the umask, are you referring to the `mode` (2nd) parameter of `proc_create_data()`? There are 2 types of files, those to be written, and those to be read, so do you suggest we should be able to set a umask for each type?

rfuchs:
I'm not really concerned about reading state/status info, even though controlling that would also be an option. But perhaps it would make sense to allow one group to control the module rather than just one user, which would require g+w privs on the writable files. Of course you could also go nuts and have different users, groups and modes for the different types of pseudofiles etc... Anyway, it was just an idea.

camilleoudot:
The `control` pseudo files are already user and group writable, so it's just a matter of issuing the desired group id option to the module, which my patch on the init file does not do (it takes the user's effective group), but it's easy to add a separate group config value to tweak this.

camilleoudot:
Hi,
I added a parameter in `rtpengine.sysconfig` for choosing the group that owns `/proc/rtpengine`. Do you think of something else missing here?

PR #189, opened 2015-11-19
Add logging related to kernelization and strict-source flag - master branch

smititelu:
Add logging related to kernelization and strict-source flag, for master branch


rfuchs:
The _buf versions of the address printing functions (ie smart_ntop_p_buf) are available in this branch as well, if you want to update the patch to avoid having to use extra buffers everywhere.

smititelu:
Thanks. Updated pull request.

PR #202, opened 2015-12-17
Adding images to illustrate redis database layout

lbalaceanu:
Hi Richard,

If you agree with this documentation addition, I will then link the images to the https://github.com/sipwise/rtpengine/wiki/Redis-data-structure page.

Does the utils/images folder sound like a proper location for the images?


pkuzak:
The sources of the images can be modified with https://www.yworks.com/.

PR #220, opened 2016-02-12
Update connection behaviour when redis breaks down

smititelu:
Try reconnecting to redis database if it breaks down, when new actions come. Rtpengine process will continue working and not abort() anymore.


rfuchs:
OK - I suppose it's better to continue working and having RTP forwarding but a possibly stale database in case Redis disappears than having no RTP at all.

For printing Redis addresses, you can use endpoint_print_buf(&r->endpoint) instead of sockaddr_print_buf or sockaddr_print_port_buf to automatically get the port in there as well.

I think it might be possible to get rid of r->state by using the return value of redis_check_conn() - right now it always returns 0 but sets r->state - it should be possible to use the return value in place of r->state. I can do that mod afterwards myself if you'd like.

smititelu:
You are right, only the return value of redis_check_conn() could be used. The thing I'm thinking of is that without current redis state, if redis goes down, the log message "Lost connection to Redis..." will be printed every time an action comes (i.e. redis_update() upon new calls), thus being a little more harder to spot when the crash actually happened and how many times redis crashed.

IMHO, we should keep the r->state. Maybe it may be useful for future stuff also?!

rfuchs:
Fair enough, thanks.

smititelu:
Updated pull request with the latest master changes, and above comments.
Added possibility to start rtpengine even if configured redis is down. Basically tested it and it works fine.

rfuchs:
If we continue startup without connection to Redis, then no calls will be restored upon startup. I don't know about your use case, but in our case we'd need to prevent that. So this needs to be either optional or made dependent on the Redis role. What do you prefer?

pkuzak:
Good point we also had discussed about internally. How about introducing a configuration parameter like *REDIS_REQUIRED* which defaults to *yes*? If you agree, we'll update the pull request.

rfuchs:
Sure, that works for me.

smititelu:
Updated pull-request for the above comments and tested the new parameter functionality.

rfuchs:
Thanks for your work guys.

PR #222, opened 2016-02-24
Fix segfault when ps->component=0

smititelu:
@pkuzak discovered a scenario which constantly lead to rtpengine coredump segfault:
1. Make a call with UAS answering with 183 Session Progress (rtpengine will kernelize and store stuff into Redis. Do **not** finally answer the call.
2. Restart rtpengine
3. Finally answer the call ==> Segfault and coredump

Investigating the coredump we saw that ps->component was 0 in this case, leading to a NULL sfd due to "sfd = g_queue_peek_nth(&il->list, ps->component - 1);"

Related commit: b0a38982c2eebfe55ec8b4761505d7af17774718

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -781,6 +781,7 @@ static void __assign_stream_fds(struct call_media *media, GQueue *intf_sfds) {
 			il = l->data;
 
 			sfd = g_queue_peek_nth(&il->list, ps->component - 1);
+			if (!sfd) return ;
 
 			sfd->stream = ps;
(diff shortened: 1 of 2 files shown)


smititelu:
Updated commit with redis add and restore ps->component. Testing passed also without previous commit, but let's keep it as a failsafe.

rfuchs:
Good catch, thanks. Technically ps->component should reflect the stream's index in the linked list and can be auto-generated when building the list (which is what __num_media_streams does). But I suppose it doesn't hurt to retain it explicitly.

PR #225, opened 2016-03-01
Redis keyspace notifications

lbalaceanu:
This pull request introduces changes that allow for a rtpengine with configured redis support to receive redis generated notifications from multiple keyspaces:

    * Support for 1 write and multiple read/listen keyspaces
    * Listen-Keyspaces can be configured via configuration file (SUBSRIBE_KEYSPACES)
    * Listen-Keyspaces can be added/deleted/listed via rtpengine-ctl: ksrm, ksadd, kslist

On receiving notification for call creation through redis, a new structure is created inside rtpengine which can serve the connection in user space.
This feature does not interfere with normal rtpengine operation (not using SUBSRIBE_KEYSPACES config parameter or the mentioned rtpengine-ctls means rtpengine behaves normally)
Statistics have been changed to only refer to calls created via command channel ("own" sessions)


rfuchs:
Just a quick note to let you know that this will take me a while to go through, especially since I'll be away next week.

pkuzak:
Sure, thanks for that note. Though this feature already works pretty well (at least for our use case), we're yet probably not completely done with it. So chances are high that we'll come up with more pull requests in the future improving this feature. Nonetheless we thought it's better to create a rather slim pull request so it's easier to review.

rfuchs:
I've only just started going through this and I'm new to libevent, so bear with me.

First of all, the libevent docs seem to suggest that if threading (pthreads) is used, the library must be initialized with evthread_use_pthreads() and also linked against libevent_pthreads.

If I'm reading this right, the keyspaces that can be configured are Redis database numbers? This should probably be made more clear in the readme.

There seem to be a few cleanup routines missing, for example in main.c:349 a corresponding free() if the integer could not be parsed (which could actually be made a fatal error - also malloc could be avoided by storing the int as a pointer, e.g. by using GPOINTER_TO_UINT() or similar). I suppose redis_notify() should also clean up the objects it created when it exits?

I see that redisAsyncCommand() supports printf-style formats. You could use this to get rid of the extra sprintf.

onRedisNotification() expects a callmaster struct in privdata, but redis_notify_event_base_loopbreak() puts a NULL there, which would cause a segfault. I assume that the event handler is never called because event_base_loopbreak() is called first. Not sure if it can be assumed that the punsubscribe was done successfully, or if it even matters since the process will terminate afterwards anyway
terminate afterwards anyway.

I assume that event_base_dispatch() returns/breaks if there's a connection error to Redis? I don't see connection errors handled explicitly and the docs aren't clear about it.

Regarding updates to existing calls, it looks like the mechanism here is to always destroy the existing call and then re-create it using the updated details from Redis. I haven't gotten into the implications of this yet -- just to confirm that this is what's happening.

Also there seems to be a mistake in the call deletion/destruction routine. The hash lookup into cm->callhash is done without a lock, which is unsafe. The correct way is to acquire a read lock, do the lookup, increase the refcount and then release the lock (making sure to release the reference later). You should probably just use call_get() which does all that.

debian/control
```diff
@@ -7,6 +7,8 @@ Build-Depends: debhelper (>= 5),
                libcurl4-openssl-dev | libcurl4-gnutls-dev | libcurl3-openssl-dev | libcurl3-gnutls-dev,
                libglib2.0-dev (>= 2.30),
                libhiredis-dev,
+	       libevent-2.0-5,
```
rfuchs: Since these are the build dependencies, it shouldn't be necessary to put the actual binary lib as dependency here. The -dev package should be all that's required (possibly with a >= 2.0 version restriction).

pkuzak:
Thanks so far for your feedback, of course we will take care of it. 
Regarding the updates of calls, you are completely right with what you said and this is also what we've observed. Maybe you remember, @inf265 once wrote an e-Mail where we asked why it is implemented like this. Your answer was to keep the database consistent. Sure, it's a little bit annoying on a "passive" node calls getting re-created upon receiving keyspace notifies, but beside a higher amount of log lines due to this feature, from a functionality point of view it worked well and fulfilled our requirements. So we decided to leave it for now like it is. Maybe we'll take a closer look on this in the future.

We'll take care to update the pull request. Should you spot any other things meanwhile, just tell us.

smititelu:
Updated pull-request with fixes for the above comments. Changes include:
- refactor notifications event base logic
- don't alloc keyspaces integers, use GUINT_TO_POINTER/GPOINTER_TO_UINT instead
- compile and link with libevent_pthreads

rfuchs:
Looks OK for the most part. A few minor notes, but nothing major.

Just to confirm that this is what's intended, when notifications are enabled, two connections to Redis are established, one "async" for notifications and one regular one for actually reading from Redis?

guint_cmp is obsolete since the list now stores the numbers directly. g_queue_find() can be used instead of g_queue_find_custom()

cli_incoming_terminate() is missing locking for the callhash when iterating through the calls. cli_incoming_ksrm() has the same issue.

conf->redis_subscribed_keyspaces should also be protected by conf->config_lock wherever it's accessed and modified.

There still seems to be some cleanup routines missing. The loop in redis_notify_loop() repeatedly calls redis_async_context_alloc() but the objects allocated there (in particular the redis_notify_async_context) are never freed within that loop.

Can you explain the addition of the --ignore-missing-info parameter to dh_shlibdeps? The package seems to build fine for me without it.

Thank you

smititelu:
Right now, the redis_notify "sync" context is used to regularly ping the redis server and check for availability, each 1 second after the connection was lost (because the tread blocks on event_base_dispatch() and unblocks if connection was lost). The "async" context is used for notifications only. A future improvement to this might be not using the "sync" context at all, but use some sort of "async" time callback to check the connection (http://stackoverflow.com/questions/13568465/reconnecting-with-hiredis).

I was also thinking of a redis_async_context_free(), but I was getting a segfault with "Cannot access memory at .." each time. After looking at the hiredis README I found that the async context is freed after intended or unintended disconnection, after disconnect callback is fired.(https://github.com/redis/hiredis#asynchronous-api).

Asking @inf265 about --ignore-missing-info, he said that he had problems building rtpengine on his (older) ubuntu and that line solved it. I tried without it also and it builds fine(Ubuntu 14.04.4 LTS, Debian 8); I'll remove the line.

Thank you for your review. Will update the pull-request soon.

rfuchs:
Thank you for the explanations.

smititelu:
I am having some trouble keeping the read hashlock while iterating through the callhash, on ksrm and terminate, mainly because `call_destroy()` takes the write hashlock inside; only stacking multiple read locks by a single thread is defined by pthreads, the other combinations being undefined behaviour(I actually tried the undefined behaviours and they worked with no deadlocks, but still..)

If I try to take the read lock and release it right before `call_destroy()` it could happen that the call is meanwhile being destroyed by other means(e.g. due to BYE or timeout) and even though there is `c != NULL`, the memory at that location could be freed already, leading to an invalid memory location access. Other idea would be to build a queue of calls to be delete, protected by the read hashlock while iterating through the callhash and then just destroy the calls in that queue; but this won't solve the previous problem either. Do you have some proposals for this issue?

Also, I'd like to ask you if it would be ok to add the `obj_put(call)` right at the end of `call_destroy()` function. Thus it won't be needed to call it every time after `call_destroy()`. Do you think this would have major, unwanted, side-effects?

rfuchs:
The procedure is to acquire the read lock, access the hash, put a pointer to a call into a local variable, increase the reference count (obj_get/obj_hold) and then release the read lock. This guarantees that the call won't be freed until you're done with it and you release your reference (obj_put).

Regarding obj_put within call_destroy, I see a few instances of call_destroy being called where the calling function doesn't hold an explicit reference to the call, so that wouldn't work. However you're welcome to add a short wrapper (like a 2-liner inline function) which does both.

rfuchs:
Since we're on the subject of wrappers - if iterating through the list of calls becomes a common task, then we could even have a wrapper function for doing this, i.e. something which does all the locking and refcounting and then calls a callback function. I know there's at least one other instance in call.c (callmaster_timer). This one takes a different approach in that it holds the read lock for the entire duration of the hash walk and instead builds a list of calls to destroy, which is then iterated after the hash lock has been released. The benefit of this approach is that it avoids restarting the hash walk each time a call is destroyed.

smititelu:
Thank you for the tips. Now the actual purpose of obj_* functions is clear to me. Therefore I propose the following 2 commits.

I've basically tested the changes without spotting any issues. Further tests will follow with respect to memory consumption; I just want to ask for you opinion on the latter code changes.

rfuchs:
These changes look fine now. Refcounting and locking logic seems OK.

Only thing I noticed was that destroy_own_foreign_calls() doesn't free the call_list. You can use the following snippet to free the elements as you iterate through the list:

```
while ((c = g_queue_pop_head(&call_list))) {
```

This works since `c` (`l->data`) is guaranteed not to be NULL by the code just above. Alternatively you can use g_queue_clear() at the end of the function.

daemon/cli.c
```diff
@@ -716,6 +717,7 @@ static void cli_incoming_terminate(char* buffer, int len, struct callmaster* m,
 
    call_destroy(c);
    obj_put(c);
+   log_info_clear();
```
rfuchs: Consider putting log_info_clear() right at the end of cli_incoming(). It's always safe to call it and this would handle any other cases of log_info refcount issues that could arise in the future. That would put it right before returning to the main loop, which is where it should go.

smititelu:
Refcounting works ok. Currently testing how system memory is behaving.

smititelu:
Even if the refcounting logs look ok, memory leaks still happen. Running with Vlagrind it gives me this [1]. At a quick glance on the code, I think the redis reply object should be freed (`freeReplyObject()`), in the `redis_build_list_cb()`. I will give a try with the changes.

[1] http://pastebin.com/YBy9xmYT

rfuchs:
I think you're right. redis_restore() also seems to be missing that.

smititelu:
After solving the previous leaks and testing (6k calls, 100 calls/sec) I could see an improvement. However, one can still see system memory leaks. Valgrind reports [1] for the above scenario. IMHO, this can happen only when links between tags and maps/medias can't be established, so `__call_free()` won't free them.

What I'm thinking of is not using `call_destroy()` in `redis_restore_call()` but "manually" free the allocated structures in case of failures. It might be tricky to synchronize the callhash content if errors happen on SADD(here)->DEL->SADD notification (to ensure that `__call_free()` won't be called anymore which will lead to segfault). What do you think? Do you have any other ideas to solve the latter Valgrind reports.

[1] http://pastebin.com/BdsgmwEj

rfuchs:
Let's see. These two leaks are from uid_slice_alloc0() in redis_medias() and redis_maps(). Looking at the code, I think the problem is an inconsistency between how the lists are built and how __call_free() operates.

uid_slice_alloc0() uses a GQueue in the main `call` struct to generate and keep its unique indexes. It allocates the struct and adds it to the queue given in one go and from that it derives its unique index. So technically, every struct is linked into the `call` and so should be able to be freed.

However, __call_free() doesn't operate on these queues. Instead, it frees the `media` structs by walking through the list of `monologue` structs and in turn frees the `endpoint_map` structs by walking through the `media`. But both of these lists aren't populated until later in the restore process, so if there's an error somewhere in between, they remain empty.

The solution would be to reorder the process in __call_free() to make it operate in the main queues in the `call` struct, as those are guaranteed to hold all the data. All it should require is to switch the logic between `call->medias` and `monologue->medias` around, and the same for the `endpoint_map`

smititelu:
With the latest 2 commits, the memory looks stable on both active(calls terminated via BYE)/passive(calls terminated via DEL) machines.

daemon/call.c
```diff
@@ -208,6 +205,12 @@ static void call_timer_iterator(void *key, void *val, void *ptr) {
 		goto delete;
 	}
 
+	// other timeouts not applicable to foreign calls
+	if (IS_FOREIGN_CALL(c)) {
+		c->deleted = c->ml_deleted = poller_now + cm->conf.delete_delay;
```
rfuchs: I'm wondering about this addition. This always marks all foreign calls as deleted and keeps moving the deletion timestamp into the future (or into the present, as delete_delay can be 0). What's the reasoning behind this?

daemon/aux.h
```diff
@@ -82,8 +82,7 @@ unsigned int in6_addr_hash(const void *p);
 int in6_addr_eq(const void *a, const void *b);
 unsigned int uint32_hash(const void *p);
 int uint32_eq(const void *a, const void *b);
-
-
+int guint_cmp(gconstpointer a, gconstpointer b);
```
rfuchs: Leftover prototype - should be removed

daemon/call.c
pkuzak: The idea behind this was: Foreign calls do not have media in normal operation mode (no redundancy fail-over scenario). In this case, we didn't want _rtpengine_ to delete this foreign calls due to _TIMEOUT_ or _SILENT_TIMEOUT_. Marking foreign calls as deleted seems to be from a time before we've introduce the _foreign_ flag. Now I think it should be enough to just check if a call is a foreign one or not, so no need marking foreign calls as deleted. We'll update the pull request. Ideas are welcome.

smititelu: As @pkuzak said, that part of the code is a remnant from an old idea. From what I remember, the idea was that if a rtpengine machine became responsible for a foreign call (i.e. started receiving media for that call) to also consider the DELETE_DELAY from the point which no media was being received. But we decided to keep it simple and terminate foreign calls only via normal DEL notifications, FINAL_TIMEOUT and cli commands.

PR #227, opened 2016-03-07
Confirm also the sink when ASYMMETRIC flag is used

smititelu:
The call scenario we are facing is is similar to:
* RTP stream UAC --> UAS goes over rtpengine1
* RTP stream UAS --> UAC goes over rtpengine2

Need to kernelize the call legs even if asymmetric flag is used.


rfuchs:
Hm, I'm not sure if this is the correct solution to your scenario. The 'asymmetric' flag is supposed to handle a different scenario (source IP/port endpoints aren't the same as the destination ones).

Are you saying that you have two rtpengine instances, both handling the same call, and RTP flowing A -> B is handled by one instance, but B -> A by the other?

smititelu:
> Are you saying that you have two rtpengine instances, both handling the same call, and RTP flowing A -> B is handled by one instance, but B -> A by the other?

Yes, this is the scenario we have. We are using flags "asymmetric trust-address ICE=remove replace-origin strict-source via-branch=extra".

The streams don't get kernelized if asymmetric flag is used, due to "sink not CONFIRMED" error. The error does not appear anymore if the asymmetric flag is removed.

rfuchs:
I can't really reproduce this. With unidirectional RTP, the stream isn't kernelized either way, with or without the asymmetric flag present.

Since the asymmetric flag isn't supposed to address this use case, I'd propose an additional flag, say `unidirectional` to do it. I can do the implementation if you'd like.

pkuzak:
Thank you for your feedback and the proposed idea. As we can easily reproduce and test this, I'd say we can also take care about the implementation. We'll update the pull request afterwards so you can have a review.

smititelu:
Updated the pull-request with the `unidirectional` flag.

PR #229, opened 2016-03-08
Add FINAL_TIMEOUT parameter

smititelu:
The timeout that will limit the duration of a call.


rfuchs:
Just a quick question here. When checking the timeout, you take c->last_signal into account, which means that reinvites would keep the call alive past the timeout. Is this intended?

smititelu:
It is intended to check the call since creation time. Thanks for noticing this.
I've updated pull request also adding rtpengine-ctl setters/getters for this params.

rfuchs:
Well, there still is concurrency between reading threads and writing threads, so this must be taken into account. Normally reads and writes to integers are atomic and these values aren't sensitive enough to warrant a lock, but this isn't guaranteed and in some cases (different architectures) accesses might not be atomic, which would open the door to a race condition. So better be on the safe side. Either keep the lock or use dedicated atomic ops (g_atomic_* for example) to access the values.

smititelu:
I've added a `rwlock_t` in `callmaster_config` structure, in commit 2830e42. I think it should solve the concurency between reading and writing threads. I'm not 100% percent if you've seen it :D. Is your above comment still applicable after that commit?

rfuchs:
Yes I've seen it, thank you. My comment was meant as an explanation of the situation and to offer an alternative solution. But the code is fine as it is now.

rfuchs:
Actually, I take that back. :) There's one thing missing. The config_lock is never initialized (rwlock_init)

smititelu:
I've added the the rwlock_init. Thank you for noticing.

rfuchs:
Slight complication there. :) The conf structure is overwritten in main.c:610, so the init would have to be done elsewhere, e.g. main.c:531.

smititelu:
I've put it in main.c:537 right after mc gets ZERO'ed.

PR #231, opened 2016-03-14
Add offer/answer/delete processing statistics

smititelu:
Add graphite offer/answer/delete min/max/avg statistics.
Print new graphite statistics in rtpengine-cli list totals.

### DIFF
--- a/daemon/aux.h
+++ b/daemon/aux.h
@@ -567,6 +567,11 @@ INLINE void timeval_multiply(struct timeval *result, const struct timeval *a, co
 	timeval_from_us(result, timeval_us(a) * multiplier);
 }
 INLINE void timeval_divide(struct timeval *result, const struct timeval *a, const long divisor) {
+	if (divisor == 0) {
+		result->tv_sec = 0;
+		result->tv_usec = 0;
+		return ;
+	}
 	timeval_from_us(result, timeval_us(a) / divisor);
 }
 INLINE void timeval_add(struct timeval *result, const struct timeval *a, const struct timeval *b) {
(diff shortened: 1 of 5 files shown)


rfuchs:
I think you might be able to greatly reduce the amount of code (and especially redundant code) by unifying the variables into a struct, e.g.

```
struct foobar {
  mutex_t lock;
  u_int64_t count;
  struct timeval time_min, time_max, time_avg;
};
...
  struct foobar offer, answer, delete;
```

and then move the selection of the appropriate struct to the calling function, e.g.

```
static void timeval_update(struct foobar *foo, const struct timeval *diff) {
  mutex_lock(&foo->lock);
...
}
...
  timeval_update(&c->callmaster->totalstats_interval.offer, &stop_tv);
```

What do you think?

smititelu:
You are right. The code can be simplified a lot by doing this.
I've updated the pull request accordingly.

daemon/aux.h
```diff
@@ -567,6 +567,7 @@ INLINE void timeval_multiply(struct timeval *result, const struct timeval *a, co
 	timeval_from_us(result, timeval_us(a) * multiplier);
 }
 INLINE void timeval_divide(struct timeval *result, const struct timeval *a, const long divisor) {
+	if (divisor == 0) return ;
```
rfuchs: This is causing compiler warnings. If divisor is 0 then *result is left uninitialized which may then end up being used for other calculations further down the pipe (e.g. in timeval_totalstats_average_add())

smititelu:
Updated pull-request to return (0 sec, 0 usec) timeval in case divisor is 0.

PR #234, opened 2016-04-11
Update obj_* debugging

smititelu:
We are investigating memory leaks that happen **only** on machines which establish calls via redis notifications (see #225). Leaks are not happening on machines that establish calls via offer/answer.

We believe leaks happen in rtpengine `redis_restore_call()`. Possible causes:
- structures are not freed due to refcounting (this commit will help in debugging this)
- free callback function of structures does not free all the content of calls established via `redis_restore_call()`; however, they free all the content of calls established via `offer/answer` since the leak can't be noticed there.
- redis_restore_call() leak memory inside (e.g. forget to free the content of a glib queue etc..); this can't be noticed on current behaviour since `redis_restore()` is called only when rtpengine starts. But when notifications are turned on, it will receive SADD->DEL->SADD for each new foreign call which translates to (`redis_restore_call()`->`call_destroy()`->`redis_restore_call()`).

An idea, to be 100% sure that the leaks come from redis_restore_call(), is to have a rtpengine-ctl command that reloads redis calls on-demand. Thus, by repeatedly doing 'reload + terminate all' one can watch the system memory behaviour. Also I will investigate more using Valgrind.

What do you think about the possible causes?
What do you think about having a redis reload ctl command?

Other ideas are welcomed.

### DIFF
--- a/daemon/obj.h
+++ b/daemon/obj.h
@@ -43,27 +43,27 @@ struct obj {
 
 #define OBJ_MAGIC 0xf1eef1ee
 
-#define obj_alloc(t,a,b)	__obj_alloc(a,b,t,__FILE__,__LINE__)
-#define obj_alloc0(t,a,b)	__obj_alloc0(a,b,t,__FILE__,__LINE__)
-#define obj_hold(a)		__obj_hold(&(a)->obj,__FILE__,__LINE__)
-#define obj_get(a)		__obj_get(&(a)->obj,__FILE__,__LINE__)
-#define obj_put(a)		__obj_put(&(a)->obj,__FILE__,__LINE__)
-#define obj_hold_o(a)		__obj_hold(a,__FILE__,__LINE__)
-#define obj_get_o(a)		__obj_get(a,__FILE__,__LINE__)
-#define obj_put_o(a)		__obj_put(a,__FILE__,__LINE__)
+#define obj_alloc(t,a,b)	__obj_alloc(a,b,t,__FILE__,__func__,__LINE__)
+#define obj_alloc0(t,a,b)	__obj_alloc0(a,b,t,__FILE__,__func__,__LINE__)
+#define obj_hold(a)		__obj_hold(&(a)->obj,__FILE__,__func__,__LINE__)
+#define obj_get(a)		__obj_get(&(a)->obj,__FILE__,__func__,__LINE__)
+#define obj_put(a)		__obj_put(&(a)->obj,__FILE__,__func__,__LINE__)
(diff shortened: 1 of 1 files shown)


rfuchs:
Hi,

To clarify some of the obj_ functions used here.

The _o versions of the functions are designed to work directly on `struct obj` type variables, while the regular (non-_o) versions work on the embedded `struct obj obj` members found in other structs. The __obj versions aren't supposed to be used directly and I see that this is an error in log.h. I'll fix this separately, which will cause a conflict with this pull request - you can just remove those sections then.

I'm OK with the other changes here, minus enabling the OBJ_DEBUG flag - I'd prefer to leave this disabled in the git repo and then enable it only locally.

I'm also OK with an additional -ctl reload function if it helps you. Perhaps use #ifdefs to include it only in debug builds, so people don't accidentally use it on production systems?

The issue you're having is almost certainly related to a refcount booboo somewhere. In case you're not aware, you need to set G_SLICE=always-malloc when debugging with valgrind to defuse the glib slice allocator. Let me know if you need help hunting down the ref leak. The obj debugging stuff enabled should definitely be helpful.

smititelu:
Your intuition is right. I am sure that the memory leak is a **call** refcount issue. See http://pastebin.com/FJca6vUq.

Thanks for your advices. I'll let you know if I have some troubles in tracking it down.

rfuchs:
0206f09 contains the fixed obj debugging stuff.

Just a quick question, do you see the ref/mem leak only for calls managed through the new notification interface, or for all calls restored from redis?

rfuchs:
What you see in the snipped that you pasted is probably because `call_get() also uses the call for `log_info` (through `log_info_call()`) which holds an additional reference to the call. It gets released when `log_info` is set to something else or explicitly released through log_info_clear()

Thus, a call_get*() needs to be paired not only with a obj_put() but also with a log_info_clear().

rfuchs:
In practise, every _entry point_ into the code should have a log_info_clear() at its end. Entry points are event handlers, such as control_ng_incoming() or stream_fd_readable(). They should use log_info_clear() just before returning to the handler loop.

smititelu:
For current upstream master, for calls established via `offer/answer` or via `redis_restore` this is applicable:
- call terminate via BYE => refcounting ok
- call terminate via `terminate all` => refcounting ok
- call terminate via `terminate call_id` => refcounting **not** ok; struct call is not freed. After calling `log_info_clear()' right before obj_put(),  struct call refcounting is ok.

Thank you for your notes here; it helped alot. I will close this PR and continue working on the notifications PR, considering everything from here.

PR #239, opened 2016-05-11
Add >= idx check for redis_list_get_idx_ptr()

smititelu:


smititelu:
We are experiencing a segfault only on backup machines (i.e. FOREIGN calls, restored via redis) when printing the CDR upon `call_destroy()`:
```
(gdb) bt
#0  call_destroy (c=c@entry=0x7fad110fb960) at call.c:2045
#1  0x0000000000417c8f in onRedisNotification (actx=<optimized out>, reply=0x7fad006fec80, privdata=0x2566980) at redis.c:339
#2  0x00007fad2ba0d100 in redisProcessCallbacks () from /usr/lib/x86_64-linux-gnu/libhiredis.so.0.10
#3  0x00007fad2c8d5254 in event_base_loop () from /usr/lib/x86_64-linux-gnu/libevent-2.0.so.5
#4  0x0000000000418599 in redis_notify (cm=cm@entry=0x2566980) at redis.c:543
#5  0x00000000004189f8 in redis_notify_loop (d=0x2566980) at redis.c:587
#6  0x000000000040bc0f in thread_detach_func (d=0x2556d40) at aux.c:143
#7  0x00007fad2b7f0b50 in start_thread () from /lib/x86_64-linux-gnu/libpthread.so.0
#8  0x00007fad2b53b30d in clone () from /lib/x86_64-linux-gnu/libc.so.6
#9  0x0000000000000000 in ?? ()

(gdb) print md
$5 = (struct call_media *) 0x7fad027a9540
(gdb) print md->index
$6 = 1
(gdb) print protocol
$7 = 0x43ea5f "rtp"
(gdb) print cdrlinecnt
$10 = <optimized out>
(gdb) print addr
$11 = 0x7fad2706f5e4 "91.60.65.152"

(gdb) print ps
$8 = (struct packet_stream *) 0x7fad050555f0
(gdb) print ps->stats
$9 = {packets = {p = 0x1}, bytes = {p = 0xac}, errors = {p = 0x0}, delay_min = 0, delay_avg = 0, delay_max = 0, in_tos_tclass = 0 '\000', foreign_sessions = {p = 0x0}}
(gdb) print ps->selected_sfd
$4 = (struct stream_fd * volatile) 0x50
(gdb) print ps->selected_sfd->socket
Cannot access memory at address 0x70
(gdb) print ps->selected_sfd->unique_id
Cannot access memory at address 0x68
(gdb) print ps->selected_sfd->call
Cannot access memory at address 0xc8
```

So far, the only reason/solution we have is this pull-request (not 100% confirmed; just trying to traceback the code). Are there any other ideas why ps->selected_sfd would point to that `0x50` address?!

rfuchs:
So you happen to still have the raw data from Redis for this particular call?

smititelu:
Nope.. I can't find redis content for it. Probably it get deleted on all machines by the delete()/DEL notifications that triggered also this call_destroy().

smititelu:
A curious stuff found:
```
(gdb) print (((struct call*)c)->stream_fds)
$46 = {head = 0x3ac46a0, tail = 0x7fad11fa0c00, *length = 8*}
(gdb) print ((struct call*)c)->streams
$24 = {head = 0x5c106e0, tail = 0x3e53940, *length = 8*}
```

```
(gdb) print ((struct stream_fd *)(((struct call*)c)->stream_fds.head)->next->next->data)
$68 = (struct stream_fd *) 0x7fad0046a000
(gdb) print ((struct stream_fd *)(((struct call*)c)->stream_fds.head)->next->next->next->data)
$67 = (struct stream_fd *) 0x7fad0156fbd0
```
but:
```
(gdb) print ((struct stream_fd *)(((struct call*)c)->stream_fds.head)->next->next->data)->stream->selected_sfd
$65 = (struct stream_fd * volatile) 0x50
(gdb) print ((struct stream_fd *)(((struct call*)c)->stream_fds.head)->next->next->next->data)->stream->selected_sfd
$66 = (struct stream_fd * volatile) 0x0
```

For all other 6 sfds the values matched.

The scenario is one of a audio failure branching in which 4 monologues are created in total. The kamailio commands to active rtpengine are like this:
* offer1 - triggers SADD notification (because there is nothing to DEL)
* answer1 - triggers DEL+SADD notification
* offer2 - triggers DEL+SADD notification
* answer2 - triggers DEL+SADD notification
* delete - triggers DEL notification
* offer3 - triggers SADD notification (because there is nothing to DEL)
* answer3 - triggers DEL(here we had a crash) + SADD
* delete - triggers DEL notification (and clears the redis database)

A redis dump would have been very useful and have given a clue how the structures look in redis and how are linked upon restoring call; but I think we won't be able to catch it because we can not reproducing it consistently. If we could reproduce it consistently we could try to delay that `answer3` and get it.

rfuchs:
Well, the way I see it... `selected_sfd` gets assigned in `redis_link_streams`, restored from `stream-X-Y` hash's `sfd` value, which is an integer. If `selected_sfd` was NULL, then that integer was -1 and `redis_list_get_ptr` would subsequently return NULL.

Point to note that `redis_list_get_ptr` uses `redis_hash_get_unsigned`, which internally uses `strtol`, meaning that -1 is parsed correctly but is returned as a large positive integer, so that should work correctly.

`redis_list_get_ptr` in turn uses `redis_list_get_idx_ptr` which checks the index against the array length and only returns the pointer if the index is valid.

The list of `sfds` is created in `redis_get_list_hash` using the value from `num_sfds`. The `ptrs` array is alocated there but not initialized or populated. It's populated in `redis_sfds` using the array length in `sfds`. Any error results in -1 being returned, which aborts the entire restore process. I don't see any possibilities of `sfds->ptrs` being left uninitialized or initialized incorrectly when used later on.

As for `stream->selected_sfd` itself, when the stream is allocated, it's initialized to zero, so the only way it could end up as `0x50` is when it gets assigned that value explicitly, which (in the restore) can only happen in `redis_link_streams`. But it seems to be impossible for `sfds` to contain invalid pointers.

So it has to be something else. Memory corruption or concurrency would be candidates. One thing I've noticed is that `redis_restore_call` uses `call_get_or_create` to create the call struct. Originally this was guaranteed to return a blank and newly allocated call, but now this isn't necessarily true any more. Not sure if that can really make a difference though.

What were the other threads doing at the time the core was generated? (`thread apply all bt`)
What's the reference count of the call struct? (`call->obj.ref`) I believe at that point in the code it ought to be 2

lbalaceanu:
Hi Richard,

We also suspect some race condition. 
As for the threads, most are waiting as they should be ( the majority in the epoll_wait(), then 1 each for: waiting sighandler, main loop checking the g_shutdown, pthread_cond_timedwait related to ice,  sleeping waiting for graphite)

2 more are:
1 in the onRedisNotification -> call_destroy() (which causes the fault)
1 is in the call.c: callmaster_timer (line 544: atomic64add call): 
```
for (j = 0; j < ke->target.num_payload_types; j++) {
.....
    if (ke->rtp_stats[j].packets > atomic64_get(&rs->packets))
        atomic64_add(&rs->packets,
	              ke->rtp_stats[j].packets - atomic64_get(&rs->packets));
```

The reference count of the call struct is 10.

rfuchs:
Regarding that second core dump you mentioned. Does selected_sfd also point to 0x50 there, or some other value? And, does the offer_answer in the other thread pertain to the same call-id as the call that caused the segfault?

lbalaceanu:
Hi again,

This time value is 0x60:

```
(gdb) print ((struct stream_fd *)(((struct call*)c)->stream_fds.head)->next->next->data)->stream->selected_sfd
$1 = (struct stream_fd * volatile) 0x60
(gdb) print ((struct stream_fd *)(((struct call*)c)->stream_fds.head)->next->next->next->data)->stream->selected_sfd
$2 = (struct stream_fd * volatile) 0x0
```

And the call-ids (offer_answer vs segfault) don't match ( what is  more, I presume offer_answer is OWN call, while other is FOREIGN call).

Thank you

PR #245, opened 2016-06-16
Add Feature: Record call to pcap file

egreenmachine:
See README for options and how to use.


rfuchs:
Thanks - this is gonna take me a while to go through.

daemon/media_socket.c
```diff
@@ -1172,8 +1193,16 @@ static int stream_packet(struct stream_fd *sfd, str *s, const endpoint_t *fsin,
...
 	 * 1 = forward and push update to redis */
-	if (rwf_in)
+	if (rwf_in) {
 		handler_ret = rwf_in(s, in_srtp);
+	}
+
+	// If recording pcap dumper is set, then we record the call.
+	if (recording_pdumper != NULL && call->record_call) {
+		stream_pcap_dump(recording_pdumper, stream, s);
+		call->recording->packet_num++;
```
rfuchs: This seems to be insufficiently protected by a lock. Only stream-scope locks are held at this point, but it's modifying a call-scope variable. Am I right to assume that this counter is supposed to correspond to the packet index in the pcap? I don't know how thread-safe libpcap is and whether pcap_dump can be called without locking, as it seems to have the same problem. Either way, if the packet_num is supposed to correlate with the pcap packet index, then both have to be protected by the same lock, otherwise you may get out of order packets or mismatched indexes.

daemon/media_socket.h
```diff
@@ -64,6 +65,8 @@ struct stream_fd {
 	struct packet_stream		*stream;	/* LOCK: call->master_lock */
 	struct crypto_context		crypto;		/* IN direction, LOCK: stream->in_lock */
 	struct dtls_connection		dtls;		/* LOCK: stream->in_lock */
+	pcap_t				*recording_pd;
```
rfuchs: Are these here actually used?

daemon/media_socket.c
```diff
@@ -1434,7 +1462,6 @@ struct stream_fd *stream_fd_new(socket_t *fd, struct call *call, const struct lo
 	sfd->call = obj_get(call);
 	sfd->local_intf = lif;
 	g_queue_push_tail(&call->stream_fds, sfd); /* hand over ref */
-	g_slice_free1(sizeof(*fd), fd); /* moved into sfd, thus free */
```
rfuchs: Sure that this line should be removed?

egreenmachine: This looks like a mistake from my rebase onto the latest master. We originally wrote this recording stuff a few months ago. I will fix it.

daemon/media_socket.h
egreenmachine: Looks like leftover from a refactor we did. Good catch, thanks.

daemon/media_socket.c
egreenmachine: I believe you are correct here. Originally we were writing a pcap file per stream so this was safe as each stream had it's own packet counter and was locked. There is a read lock on the call's master lock at the beginning of this function, but that would be insufficient for this. I will get a fix for this going.

egreenmachine:
I made some changes that you requested. Let me know if anything else needs to be done. Thanks!

egreenmachine:
Thanks for pulling this in. Any idea when your next tagged release will be?

rfuchs:
Currently scheduled for 26th of August

PR #253, opened 2016-07-14
redundant media setup: failover due to tout on offer response

lbalaceanu:
Failover scenario because of timeout on offer response: kamailio sends offer information to rtpengine1, but responses don't make it back to kamailio.

Consider a 3 rtpengines redundant media setup. Kamailio tries to establish session with rtpengine2 even though rtpengine1 might have persisted part of the session.

rtpengine2 - deletes previous call in memory and recreates an OWN call in redis (triggers sadd notifies on rtpengine1 and rtpengine3).
rtpengine1 - receives sadd: moves call from own->foreign, deletes call (changes don't propagate in redis); create new foreign call
rtpengine3 - receives sadd:  delete call, create new foreign call

### DIFF
--- a/daemon/call_interfaces.c
+++ b/daemon/call_interfaces.c
@@ -680,6 +680,18 @@ static const char *call_offer_answer_ng(bencode_item_t *input, struct callmaster
 	if (sdp_streams(&parsed, &streams, &flags))
 		goto out;
 
+    /* Failover scenario because of timeout on offer response: siprouter tries
+     * to establish session with another rtpengine2 even though rtpengine1
+     * might have persisted part of the session. rtpengine2 deletes previous
+     * call in memory and recreates an OWN call in redis */
+	if (opmode == OP_OFFER) {
+		call = call_get(&callid, m);
+		if (call && IS_FOREIGN_CALL(call)) {
+			rwlock_unlock_w(&call->master_lock);
+			call_destroy(call);
+			obj_put(call);
+		}
+	}
 	call = call_get_opmode(&callid, m, opmode);
 
 	errstr = "Unknown call-id";
--- a/daemon/call_interfaces.c
+++ b/daemon/call_interfaces.c
@@ -680,19 +680,28 @@ static const char *call_offer_answer_ng(bencode_item_t *input, struct callmaster
 	if (sdp_streams(&parsed, &streams, &flags))
 		goto out;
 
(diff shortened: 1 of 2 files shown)


daemon/call_interfaces.c
```diff
@@ -680,6 +680,18 @@ static const char *call_offer_answer_ng(bencode_item_t *input, struct callmaster
...
+     * to establish session with another rtpengine2 even though rtpengine1
+     * might have persisted part of the session. rtpengine2 deletes previous
+     * call in memory and recreates an OWN call in redis */
+	if (opmode == OP_OFFER) {
+		call = call_get(&callid, m);
+		if (call && IS_FOREIGN_CALL(call)) {
+			rwlock_unlock_w(&call->master_lock);
+			call_destroy(call);
+			obj_put(call);
+		}
```
rfuchs: This would leak a reference if the condition is false, as "call" gets reassigned (with the same call but a new reference) below.

Do you think this check could perhaps be moved elsewhere, e.g. into call_get_or_create()?

lbalaceanu: Hi Richard,

From the first glances and without testing I think I could manage to get the code into the call_get_or_create(). However, do you think that semantically the code snippet that deletes and then creates a call would fit right in that function? If so, I will make the changes there. test and update the pull request.

Alternatively I could change the current code to have obj_put(call) be executed if call is found, irrespective of IS_FOREIGN_CALL() result.

Which alternative would you prefer?

Thank you,
Lucian

rfuchs: Good point. My primary concern is the duplicate hash lookup in the most common case. I think this can be avoided by skipping the second call_get* if the first one returned a valid call. Maybe that would be the best option.

lbalaceanu:
So I think I can enhance the code in call_interfaces.c -> call_offer_answer_ng() somehow around this form (maybe more attention to the refrences and locks) :

```c
if (opmode == OP_OFFER) {  
    call = call_get(&callid, m);
    if (call) {
        if (IS_FOREIGN_CALL(call)) {
            rwlock_unlock_w(&call->master_lock);
            call_destroy(call);
        }
    }
    else call = call_get_or_create();
}
else call = call_get(&callid, m);
```

Is this what you had in mind? We still have the hash lookup done twice in case of (OP_OFFER and new call). 

Thank you,
Lucian

rfuchs:
Yeah something like that should work. Thanks

PR #255, opened 2016-07-14
Made interactions with recording metadata threadsafe

dbmikus:
Updating, freeing, and writing the recording metadata is now threadsafe. This is in regards to the metadata that we receive from the `rtpengine_offer` or `rtpengine_answer` commands, not the `meta_fp` file.

We got an rtpengine crash that happened when calling `free(call->recording->metadata);` at https://github.com/sipwise/rtpengine/blob/master/daemon/call_interfaces.c#L753.

We are pretty sure this happened due to a double-free, but weren't able to exactly reproduce this in the code. Either way, if `call_offer_answer_ng` can be called simultaneously in multiple threads, it was technically possible to attempt to double-free the `recording->metadata` string. A mutex lock prevents that.

### DIFF
--- a/daemon/call_interfaces.c
+++ b/daemon/call_interfaces.c
@@ -749,10 +749,13 @@ static const char *call_offer_answer_ng(bencode_item_t *input, struct callmaster
 	}
 	bencode_dictionary_get_str(input, "metadata", &metadata);
 	if (metadata.len > 0 && call->recording != NULL) {
+		mutex_lock(&call->recording->metadata_lock);
 		if (call->recording->metadata != NULL) {
 			free(call->recording->metadata);
+			call->recording->metadata = NULL;
 		}
 		call->recording->metadata = str_dup(&metadata);
+		mutex_unlock(&call->recording->metadata_lock);
 	}
 	bencode_item_t *recordings = bencode_dictionary_add_list(output, "recordings");
 	if (call->recording != NULL && call->recording->recording_path != NULL) {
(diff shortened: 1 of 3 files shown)


rfuchs:
Hi, thanks for the report. Looks like you're right: call->recording->metadata isn't sufficiently protected by a lock. But I don't think it's necessary to add another mutex for this.

call->recording is supposed to be protected by call->master_lock. call->recording->metadata can then be protected either by ->master_lock also, or by ->recording->recording_lock. I think using the master_lock makes more sense. From a quick glance, it looks like all that is necessary is to move the respective code in call_offer_answer_ng up by a few lines.

Let me know what you think.

dbmikus:
@egreenmachine and I talked about using the `->recording->recording_lock`, but he brought up the point that, since `recording_lock` locks against writing RTP data to the PCAP file, updating the metadata might interfere and prevent writing output.

However, if the RTP flowing in is buffered, then I don't think we would run into issues with waiting on the lock when writing out PCAP data.

Is that correct?

If the potential problem @egreenmachine and I discussed is a non-issue, then it seems like using the `call->master_lock` and just moving the code up into that section is the solution.

rfuchs:
Either way shouldn't be a problem. During signalling (offer/answer/delete), the master_lock is always acquired in write mode, which keeps the RTP processing from acquiring a read lock, so these two will always interfere with each other in some way.

On second thought, call->recording is created and assigned with a write lock held on master_lock. The code below accesses call->recording to get to the metadata, so this must hold master_lock at least in read mode by definition. It would make most sense to simply include the metadata in what's protected by master_lock (as write lock).

rfuchs:
Looks good to me

dbmikus:
See squashed request: https://github.com/sipwise/rtpengine/pull/256

PR #300, opened 2016-12-06
Reduced complexity in onRedisNotification()

pkuzak:
If a call already exists and a Redis notification for that call is received, don't change the call (ignore notification).


rfuchs:
Seems OK, except for a small booboo which seems to be present in the original code already. The obj_put in line 1501/1474 should not be called before unlocking the master_lock (1504/1489). I'll fix it in a separate commit.

PR #302, opened 2016-12-07
Stricter build dependencies

pkuzak:
The recording feature seems to require a minimal version of libav related packages.
Tested on Wheezy and Jessie.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -4,11 +4,11 @@ Priority: extra
 Maintainer: Sipwise Development Team <support@sipwise.com>
 Build-Depends: debhelper (>= 5),
                iptables-dev (>= 1.4),
-               libavcodec-dev,
-               libavfilter-dev,
-               libavformat-dev,
-               libavresample-dev,
-               libavutil-dev,
+               libavcodec-dev (>= 6:10),
+               libavfilter-dev (>= 6:10),
+               libavformat-dev (>= 6:10),
+               libavresample-dev (>= 6:10),
+               libavutil-dev (>= 6:10),
                libcurl4-openssl-dev | libcurl4-gnutls-dev | libcurl3-openssl-dev | libcurl3-gnutls-dev,
                libevent-dev (>= 2.0),
                libglib2.0-dev (>= 2.30),
@@ -56,7 +56,7 @@ Replaces: ngcp-mediaproxy-ng
 Description: NGCP RTP/media proxy - meta package.
  This is a meta package for easy installation of all four parts of the NGCP
  media proxy. It will install the user-space daemon, the kernel-space IPtables
- module, the IPtables extension module and utility scripts. 
+ module, the IPtables extension module and utility scripts.
 
 Package: ngcp-rtpengine-kernel-source
 Architecture: all
--- a/recording-daemon/decoder.c
+++ b/recording-daemon/decoder.c
@@ -3,6 +3,7 @@
 #include <libavformat/avformat.h>
 #include <libavutil/audio_fifo.h>
 #include <libavutil/channel_layout.h>
+#include <libavutil/mathematics.h>
 #include <glib.h>
 #include <stdint.h>
 #include <libavresample/avresample.h>
@@ -378,7 +379,12 @@ void decoder_close(decoder_t *dec) {
 	if (!dec)
 		return;
 	/// XXX drain inputs and outputs
+#if LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(56, 1, 0)
 	avcodec_free_context(&dec->avcctx);
+#else
+	avcodec_close(dec->avcctx);
+	av_free(dec->avcctx);
+#endif
 	av_frame_free(&dec->frame);
 	av_frame_free(&dec->swr_frame);
 	avresample_free(&dec->avresample);
--- a/debian/ngcp-rtpengine-daemon.install
+++ b/debian/ngcp-rtpengine-daemon.install
@@ -1,3 +1,3 @@
 daemon/rtpengine /usr/sbin/
-recording-daemon/rtpengine-recording /usr/sbin/
 etc/rtpengine.sample.conf /etc/rtpengine/
+recording-daemon/rtpengine-recording /usr/sbin/
--- a/recording-daemon/mix.c
+++ b/recording-daemon/mix.c
@@ -4,6 +4,7 @@
 #include <libavfilter/buffersrc.h>
 #include <libavfilter/buffersink.h>
 #include <libavutil/channel_layout.h>
+#include <libavutil/mathematics.h>
 #include <inttypes.h>
 #include <libavresample/avresample.h>
 #include <libavutil/opt.h>


debian/control
```diff
@@ -4,11 +4,16 @@ Priority: extra
...
-               libavfilter-dev,
-               libavformat-dev,
-               libavresample-dev,
-               libavutil-dev,
+               libavcodec-dev (>= 6:10),
+               libavfilter-dev (>= 6:10),
+               libavformat-dev (>= 6:10),
+               libavresample-dev (>= 6:10),
+               libavutil-dev (>= 6:10),
+               libavcodec55 (>= 6:10) | libavcodec56,
```
mika: There are trailing white-spaces in some lines, maybe you could just drop them?
Also would be nice to run `wrap-and-sort -a`  so the changes to Build-Depends are sorted alphabetically.

pkuzak:
Thanks for commenting. I didn't know there is such a tool. I've updated this pull request, should be fine now. If not, please report back.

debian/control
```diff
@@ -4,11 +4,16 @@ Priority: extra
...
+               libavcodec-dev (>= 6:10),
+               libavfilter-dev (>= 6:10),
+               libavformat-dev (>= 6:10),
+               libavresample-dev (>= 6:10),
+               libavutil-dev (>= 6:10),
+               libavcodec55 (>= 6:10) | libavcodec56, 
+               libavformat55 (>= 6:10) | libavformat56, 
+               libavfilter4 (>= 6:10) | libavfilter5, 
+               libopus0 (>= 1.1),
+               libswscale2 (>= 6:10) | libswscale3,
```
mika: Just had a private chat with Richard regarding the Build-Dependencies and as he (absolutely correctly) pointed out e.g. the libavcodec* versions are moving forward fast so this might be annoying to maintain. Instead we (as usual) should have Build-Depends on their according -dev packages (like e.g. libavcodec-dev instead of libavcodec55 ...), or is there anything specific about that why the non-dev packages are needed too?

pkuzak:
Basically you are right and I agree with you. I think the problem is that in Wheezy, just having the depends on _dev_ packages does not work because by default it tries to install the package from the _wheezy_ repository, but it does not work with that one. We need the one from the _wheezy-backports_ repository, having version _>=6:10_ . Just adding the version to the _dev_ packages does not work, because those then depend on the corresponding _wheezy-backport_ packages, but don't get installed by default. At least we end up in build errors like: 
```
The following packages have unmet dependencies:
 libavresample-dev : Depends: libavutil-dev (= 6:10.1-1~bpo70+1) but 6:0.8.17-2 is to be installed
E: Unable to correct problems, you have held broken packages.
```
Nonetheless I've updated the pull request.

pkuzak:
Squashed commits into one. 
Regarding tuning _apt_: This is not an option for us. I am aware of this solution, but we are using a company wide build server which is not maintained by our team and most probably they will not change the configuration there. 
I haven't tried whether the build works when leaving out the versions in the build dependencies, because I wasn't able to satisfy _apt_ without them. 
I am indeed not a Debian packaging expert, but I thought this is a valid solution. 
Due to the fact that we have to use the company build server and don't have access to it to change the configuration, I am not aware of an other solution for us than proposed in this pull request. If you know a better solution, please share it with me.
You can reject this pull request if it doesn't satisfy you requirements. If we don't find an alternative, we'll have to keep this "patch" locally then (at least as long we will need to build for Wheezy).

debian/control
```diff
@@ -4,18 +4,23 @@ Priority: extra
...
-               libavfilter-dev,
-               libavformat-dev,
-               libavresample-dev,
-               libavutil-dev,
+               libavcodec-dev (>= 6:10),
+               libavcodec55 (>= 6:10) | libavcodec-dev (>= 6:11),
+               libavfilter-dev (>= 6:10),
+               libavfilter4 (>= 6:10) | libavfilter-dev (>=6:11),
+               libavformat-dev (>= 6:10),
+               libavformat55 (>= 6:10) | libavformat-dev (>=6:11),
```
guillemj: These dependencies in the style: "libavformat-dev (>= 6:10), libavformat55 (>= 6:10) | libavformat-dev (>=6:11)," are still not correct, because you are assuming private linking details of ffmpeg or libav. You might end up with a libavcodec-dev package that provides a libavcodec.so file that points to something like libavcodec.so.56, even if you have requested here libavcodec55 as Build-Depends.

If the code does not build with older packages, or you need features only available in later versions then adding a version restriction seems entirely legit, but that would be simply adding "(>= 6:10)" or similar. But again it would require verifying if this does not build with older versions. But the hardcoding of the libfooNN is just wrong here.

debian/control
```diff
@@ -4,18 +4,23 @@ Priority: extra
...
+               libavutil-dev (>= 6:10),
                libcurl4-openssl-dev | libcurl4-gnutls-dev | libcurl3-openssl-dev | libcurl3-gnutls-dev,
                libevent-dev (>= 2.0),
                libglib2.0-dev (>= 2.30),
                libhiredis-dev,
+               libopus0 (>= 1.1),
                libpcap0.8-dev | libpcap-dev,
                libpcre3-dev,
                libssl-dev (>= 1.0.1),
+               libswscale2 (>= 6:10) | libswscale-dev (>=6:11),
```
guillemj: I don't see the code making direct use of any of the stuff provided by libswscale. This dependency seems gratuitous here?

debian/control
```diff
@@ -4,18 +4,23 @@ Priority: extra
...
+               libavfilter4 (>= 6:10) | libavfilter-dev (>=6:11),
+               libavformat-dev (>= 6:10),
+               libavformat55 (>= 6:10) | libavformat-dev (>=6:11),
+               libavresample-dev (>= 6:10),
+               libavutil-dev (>= 6:10),
                libcurl4-openssl-dev | libcurl4-gnutls-dev | libcurl3-openssl-dev | libcurl3-gnutls-dev,
                libevent-dev (>= 2.0),
                libglib2.0-dev (>= 2.30),
                libhiredis-dev,
+               libopus0 (>= 1.1),
```
guillemj: And this is also wrong, as that's an implementation detail of how libavcodec was linked. If libavcodec would be linked in the past or in the future against say libopus0.0 or libopus1, then even if you explicitly depend on that it would not mean it gets used at all when linking, and less so during run-time.

recording-daemon/decoder.c
```diff
@@ -378,7 +379,12 @@ void decoder_close(decoder_t *dec) {
 	if (!dec)
 		return;
 	/// XXX drain inputs and outputs
+#if LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(56, 1, 0)
 	avcodec_free_context(&dec->avcctx);
+#else
+	avcodec_close(dec->avcctx);
+	av_free(dec->avcctx);
+#endif
```
guillemj: So, this is to be able to build with older versions of libav/ffmeg, which would seem to confirm using even older versions might indeed not be supported?

pkuzak:
So what? Shall I close this pull request or do you have an idea how to solve this without being able to change the _apt_ configuration?

rfuchs:
Restricting the -dev packages to certain minimum version numbers is not sufficient?

pkuzak:
No, because, for example, _libswscale-dev=6:10.1-1~bpo70+1_ depends on _libswscale2=6:10.1-1~bpo70+1_, but _apt_ will refuse to install it without explicitly mentioning it in the _control_ file. Similar does apply for the other _dev_ packages. We end up with errors like: 
```
The following packages have unmet dependencies:
 libavresample-dev : Depends: libavutil-dev (= 6:10.1-1~bpo70+1) but 6:0.8.17-2 is to be installed
E: Unable to correct problems, you have held broken packages.
```

guillemj:
@pkuzak I don't think this needs to be closed. There are some changes that seem mergable. The actual code changes, and perhaps the version restrictions. But the rest are just incorrect. The problem here is that the wheezy-backports repository sets the NotAutomatic and ButAutomaticUpgrades fields to "yes" which means package manager will not pull new packages not explicitly requested, it will only upgrade already installed ones (https://wiki.debian.org/RepositoryFormat#NotAutomatic_and_ButAutomaticUpgrades).

I still think this is a local setup problem that you should fix ideally in your organization. Is requesting such change to that other group managing the infra out of the question? It seems like the correct thing to do here, really.

pkuzak:
OK then. I will have a chat with the administrator of the build system describing the situation. I will report here afterwards. Though this might take a while time due to vacation time. 
So far, thanks for the review and the discussion.

pkuzak:
Works :+1: . I've updated the pull request. Now there are just dependencies on _dev_ packages. Sorry for this long discussion, but I wasn't aware of those flags.

Issue #303, opened 2016-12-07
HA: RTPEngine + Redis


markiemooster:
Hello,

I have an issue using RTPEngine and Redis in an HA setup.  I have two RTPEngine servers: one active and one standby, and Redis (+ kamailio) running on a third server.  Both RTPEngines reference the same Redis DB instance on the third server.  In addition to this, I have a WebRTC client running in Chrome that makes secure TLS SIP calls through an encrypted web socket via the active RTPEngine.  

I can successfully make calls from the client and SRTP traffic flows as it should.  When I reboot the active RTPEngine server the system correctly fails over to the other server and, after only a small pause (~1s), incoming audio continues to arrive.  However, approximately 15 seconds after this I hear nothing but silence.  

Here are the entries from the syslog of the failed over server: 

```
Dec  7 12:19:25 Web-iPath-Test-1 rtpengine[2674]: INFO: Established initial connection to Redis 172.19.1.113:6379
Dec  7 12:19:26 Web-iPath-Test-1 rtpengine[2674]: INFO: Redis restore time = 130 ms
Dec  7 12:19:26 Web-iPath-Test-1 rtpengine[2674]: INFO: Startup complete, version 4.5.2.3
Dec  7 12:19:26 Web-iPath-Test-1 rtpengine[2674]: WARNING: [3965bqul6rdtak0394f7 port 16501]: Error parsing RTCP header: invalid header version
Dec  7 12:19:26 Web-iPath-Test-1 rtpengine[2674]: INFO: Use Redis 172.19.1.113:6379 for notifications
Dec  7 12:19:26 Web-iPath-Test-1 rtpengine[2674]: INFO: Use Redis 172.19.1.113:6379 to subscribe to notifications
Dec  7 12:19:26 Web-iPath-Test-1 rtpengine[2674]: WARNING: [3965bqul6rdtak0394f7 port 16500]: Error parsing RTP header: invalid header version
Dec  7 12:19:41 Web-iPath-Test-1 rtpengine[2674]: WARNING: [3965bqul6rdtak0394f7 port 16500]: Error parsing RTP header: invalid header version
```

I'm wondering what could cause this 'invalid header version' error.  As the newly active server did not participate in the DTLS handshake where the certificates and keys were exchanged, I wondered whether this might have anything to do with it not having the client's certificate.  Does Redis store this information? 

If anyone could shed any light on this issue it would be greatly appreciated.

Thanks,
Mark

rfuchs:
[...]

PR #316, opened 2017-01-09
Implements redis onekey concept

pkuzak:
The redis onekey concepts is introduced to reduce traffic to redis
and redis notification traffic.
It modifies the current structure for one call in redis, which are
multiple keys with pre- and postfixes and the callid in between to
one key with the structure "json-<callid>". The value is a json
formatted string with the previous multi-key identifiers in it.


daemon/redis.c
```diff
@@ -1363,6 +1677,160 @@ static int redis_link_maps(struct redis *r, struct call *c, struct redis_list *m
...
+static void json_restore_call(struct redis *r, struct callmaster *m, redisReply *id, enum call_type type) {
+	redisReply* rr_jsonStr;
+	struct redis_hash call;
+	struct redis_list tags, sfds, streams, medias, maps;
+	struct call *c = NULL;
+	str s ;
+	const char *err = 0;
+	int i;
+#if !(GLIB_CHECK_VERSION(2,36,0))
+	g_type_init();
```
rfuchs: This should go into main.c around init_everything

daemon/redis.c
```diff
@@ -1363,6 +1677,160 @@ static int redis_link_maps(struct redis *r, struct call *c, struct redis_list *m
...
+#endif
+	JsonReader *root_reader =0;
+	JsonParser *parser =0;
+	char orig_json_callid[512]; memset(orig_json_callid,0,512);
+	memcpy(orig_json_callid,id->str,strlen(id->str));
```
rfuchs: This seems unnecessary if str_cut is removed since the string then isn't modified. I would suggest renaming `id` to something like `full_json_id`, then declare your own `id`
```
str id;
str_init_redis(&id, full_json_id);
str_shift(&id, strlen("json-"));
```
Which actually duplicates the `s` variable below, which then becomes redundant.

This would require updating the function signature of all called functions to receive a `str` instead of a `redisReply`.  In fact, looking at the functions, they all seem to work on the just the call-id anyway, and do receive the `call` object, so that parameter could be removed completely and the functions could just use `call->callid`. If I'm not mistaken.

daemon/redis.c
```diff
@@ -62,9 +61,23 @@ static int redisCommandNR(redisContext *r, const char *fmt, ...)
...
 #define S_LEN(s,l) (s), (size_t) (l)
+#define STRSTR(x) (x)->s
 
 #endif
 
+static void redis_restore_call(struct redis *r, struct callmaster *m, const redisReply *id, enum call_type);
+static int redis_check_conn(struct redis *r);
+static void json_restore_call(struct redis *r, struct callmaster *m, redisReply *id, enum call_type type);
+
+static int str_cut(char *str, int begin, int len) {
```
rfuchs: Looking at the instances where this function is used, I believe it should be removed. It needlessly modifies the string in place. The appropriate replacement would be to use `str_shift` or something equivalent for `redisReply` objects. I think the best way to go would be to have a `str_init_redis` function to initialize a `str` object from a `redisReply` object and the manipulate the `str` object (for example with `str_shift`). There already are a few instances of `str_init_len` in the code, used to do exactly that. They should also be replaced with a call to something like this:

```
INLINE void str_init_redis(str *out, const redisReply *rr) {
  str_init_len(out, rr->str, rr->len);
}
```

daemon/redis.c
```diff
@@ -1363,6 +1677,160 @@ static int redis_link_maps(struct redis *r, struct call *c, struct redis_list *m
...
+		rlog(LOG_DEBUG, "Could not read json data !");
+		goto err1;
+	}
+
+	c = call_get_or_create(&s, m, type);
+	err = "failed to create call struct";
+	if (!c)
+		goto err1;
+
+	c->root_reader = root_reader; // attach the json to the call in order to restore data from there
```
rfuchs: I'm not sure I like putting the root_reader into the call object like this. I understand why you're doing it since it needs to be accessible, but it doesn't really seem to belong into the call struct. Maybe there's a better way to achieve the same thing.

daemon/redis.c
```diff
@@ -1582,7 +2058,66 @@ int redis_restore(struct callmaster *m, struct redis *r) {
...
+{
+	char tmp[2048]; ZERO(tmp);
+
+	if (!p->crypto_suite)
+		return -1;
+
+	sprintf(tmp,"%s-crypto_suite",key);
+	json_builder_set_member_name (builder, tmp);
+	ZERO(tmp);
+	json_builder_add_string_value (builder, p->crypto_suite->name);
```
rfuchs: There's a lot of code repetition here and below. Always sprintf, _set_member_name and _add_*_value. It should be possible to remove the code duplication using wrapper functions or perhaps macros. I'm surprised glib doesn't provide more convenient functions. I believe this is why I've always used json-c so far :)

daemon/redis.c
```diff
@@ -1582,7 +2058,66 @@ int redis_restore(struct callmaster *m, struct redis *r) {
...
+	json_builder_set_member_name (builder, tmp);
+	ZERO(tmp);
+	snprintf(tmp,sizeof(p->master_salt), "%s",p->master_salt);
+	json_builder_add_string_value (builder, tmp);
+	ZERO(tmp);
+
+	sprintf(tmp,"%s-unenc-srtp",key);
+	json_builder_set_member_name (builder, tmp);
+	ZERO(tmp);
+	sprintf(tmp,"%i",p->session_params.unencrypted_srtp);
```
rfuchs: I suppose integer values are sprintf'd into strings and then stored as strings (instead of storing them as json integers) because the restore functions expect strings and then parse them into integers?

daemon/redis.c
```diff
@@ -1645,6 +2220,590 @@ static void redis_update_dtls_fingerprint(struct redis *r, const char *pref, con
...
+#if !(GLIB_CHECK_VERSION(2,36,0))
+	g_type_init();
+#endif
+	JsonBuilder *builder = json_builder_new ();
+
+	char tmp[2048]; ZERO(tmp);
+
+	json_builder_begin_object (builder);
+	{
+		sprintf(tmp,"json-%s",STRSTR(&c->callid));
```
rfuchs: The appropriate way to use a `str` in a printf-like format is with `STR()` and `STR_FMT`, like this:
`sprintf(tmp,"json-" STRFMT, STR(&c->callid));`
The `STRSTR` macro should probably be removed as there's no guarantee the string is null terminated. (In these cases, the strings actually should all be null terminated, but it's better not to trust this.)

inf265:
Hi Richard,
I write you back since this stuff is mostly my implementation :-)
Thank you very much for the comments on this. I'll take care and fix the your mentioned points.
Cheers, Frédéric

pkuzak:
@rfuchs , regarding the _root_reader_, could we go with a trade-off and keep it like it is in this pull request (being part of the call object)? We are going to refactor the code anyway to fit better with the new implementation, so we would fix this in a future pull request. 
We've put the _root_reader_ to the call object because we didn't find a solution how to achieve this in a different way without introducing bigger changes.

daemon/redis.c
inf265: STR_FORMAT ist meant instead of STR_FMT, right ? Like:
sprintf(tmp,"json-" STR_FORMAT,STR(&c->callid));

daemon/Makefile
```diff
@@ -26,10 +27,13 @@ LDFLAGS+=       `pkg-config --libs libpcre`
 LDFLAGS+=       `pkg-config --libs libcrypto`
 LDFLAGS+=	`pkg-config --libs openssl`
 LDFLAGS+=	`pkg-config --libs libevent_pthreads`
+LDFLAGS+=	`pkg-config --libs`
```
inf265: Deleted.

daemon/Makefile
```diff
@@ -26,10 +27,13 @@ LDFLAGS+=       `pkg-config --libs libpcre`
...
 LDFLAGS+=	-lpcap
 LDFLAGS+=	`pcre-config --libs`
 LDFLAGS+=	`xmlrpc-c-config client --libs`
 LDFLAGS+=	-lhiredis
+LDFLAGS+=       `pkg-config --libs libcurl`
```
inf265: Deleted.

daemon/main.c
```diff
@@ -298,6 +299,7 @@ static void options(int *argc, char ***argv) {
 		{ "redis-write",'w', 0, G_OPTION_ARG_STRING,    &redisps_write, "Connect to Redis write database",      "[PW@]IP:PORT/INT"       },
 		{ "redis-num-threads", 0, 0, G_OPTION_ARG_INT, &redis_num_threads, "Number of Redis restore threads",      "INT"       },
 		{ "redis-expires", 0, 0, G_OPTION_ARG_INT, &redis_expires, "Expire time in seconds for redis keys",      "INT"       },
+		{ "redis-multikey", 0, 0, G_OPTION_ARG_INT, &redis_multikey, "Use multiple redis keys for storing the call (old behaviour) DEPRECATED", "INT" },
```
inf265: Changed to G_OPTION_ARG_NONE

daemon/redis.c
```diff
@@ -312,23 +317,41 @@ void onRedisNotification(redisAsyncContext *actx, void *reply, void *privdata) {
...
 	}
 
-	pch += strlen("notifier-");
 	str_cut(rr->element[2]->str,0,pch-rr->element[2]->str);
 	rr->element[2]->len = strlen(rr->element[2]->str);
 	rlog(LOG_DEBUG,"Redis-Notifier:%s:%d: Processing call with callid: %s\n", rr->element[3]->str, r->db, rr->element[2]->str);
 
-	str_init(&callid,rr->element[2]->str);
+        // strip off json- prefix from callid
+	str_init_len(&callid,rr->element[2]->str+strlen("json-"),strlen(rr->element[2]->str)-strlen("json-"));
```
inf265: Fixed.

daemon/redis.c
```diff
@@ -719,7 +772,90 @@ static void redis_delete_call(struct call *c, struct redis *r) {
 	redis_consume(r);
 }
 
+// stolen from libhiredis
+extern redisReply *createReplyObject(int type);
```
inf265: Removed and made the function with macro INLINE

daemon/redis.c
```diff
@@ -719,7 +772,90 @@ static void redis_delete_call(struct call *c, struct redis *r) {
...
+		sprintf(key_concatted, "%s-%s-%u",key,which->str,id);
+	}
+
+
+	if (!json_reader_read_member (c->root_reader, key_concatted)) {
+		rlog(LOG_ERROR, "Could not read json member: %s",key_concatted);
+		goto err;
+	}
+
+	out->ht = g_hash_table_new_full(g_str_hash, g_str_equal, free_func, free_func);
```
inf265: wrapper omitted

daemon/redis.c
inf265: moved to main

daemon/redis.c
inf265: see comment from @pkuzak

daemon/redis.c
rfuchs: Yes, sorry. In fact it should be `sprintf(tmp,"json-" STR_FORMAT, STR_FMT(&c->callid));` The `STR` macro is to be used in redis_pipe calls.

daemon/redis.c
```diff
@@ -719,7 +772,90 @@ static void redis_delete_call(struct call *c, struct redis *r) {
...
+		unsigned int id)
+{
+	char key_concatted[256];
+	memset(key_concatted,0,256);
+
+	if (!c)
+		goto err;
+
+	if (id == -1) {
+		sprintf(key_concatted, "%s-%s",key,which->str);
```
inf265: done

daemon/redis.c
inf265: introduced makros for omitting code repetition

daemon/redis.c
inf265: Yes I'd like to do the conversions myself

daemon/redis.c
inf265: Hi Richard,
I totally agree with you but: I had a look bout a quick change and changing this would lead to work which would be for nothing in the future since I guess Pawel already said that we would like to remove the multikey stuff and keep the json only. As a consequence, all the redisreply stuff would be removed there, as well as the "notifier-" thing. Modifying the string in place was the easiest way to have a running result with the drawback of having a code situation in between the multikey and the json. I really like to suggest to keep it and with the cleanup of the multikey these things will be solved with that. What du you think ?

rfuchs:
Hi, I see you pushed the fixes to a different branch. Can you please update the pull request (squash/amend commit preferred if possible) so I can look over them again, thanks.

metzico:
Ok sorry. I rebased into redis-onekey-concept-master

inf265:
Ok sorry I rebased into redis-onekey-concept-master

daemon/redis.c
```diff
@@ -876,6 +1030,37 @@ static void *redis_list_get_ptr(struct redis_list *list, struct redis_hash *rh,
...
+
+static int json_build_list_cb(GQueue *q, struct call *c, const char *key, const str *callid,
+		unsigned int idx, struct redis_list *list,
+		int (*cb)(str *, GQueue *, struct redis_list *, void *), void *ptr)
+{
+	str s;
+	char key_concatted[256];
+	memset(&key_concatted,0,256);
+
+	sprintf(key_concatted, "%s-%s-%u",key,callid->s,idx);
```
rfuchs: This should also be `snprintf`. The `memset` above can be omitted.

daemon/redis.c
```diff
@@ -1581,8 +2055,48 @@ int redis_restore(struct callmaster *m, struct redis *r) {
 	return ret;
 }
 
+#define JSON_SET_NSTRING(a,b,c,d) do { \
+		sprintf(tmp,a,b); \
+		json_builder_set_member_name (builder, tmp); \
+		ZERO(tmp); \
+		snprintf(tmp,sizeof(d), c,d); \
```
rfuchs: I believe this should be `sizeof(tmp)` instead of `sizeof(d)`. The above call to `sprintf` should also be protected with `snprintf` in the same way. The calls to `ZERO` can be omitted. There's no reason not to use `snprintf` everywhere here (including the few more instances further down), so `JSON_SET_STRING` should be removed completely. Sorry for being picky about this.

daemon/redis.c
inf265: changed

daemon/redis.c
inf265: changed

daemon/redis.c
```diff
@@ -283,13 +280,22 @@ void onRedisNotification(redisAsyncContext *actx, void *reply, void *privdata) {
...
+			rlog(LOG_ERROR,"Redis-Notifier: The prefix 'notifier-' to determine the redis key has not been found in the redis notification !\n");
+			goto err;
+		}
+		pch += strlen("notifier-");
+	} else {
+		pch = strstr(rr->element[2]->str, "json-");
+		if (pch == NULL) {
+			rlog(LOG_ERROR,"Redis-Notifier: The prefix 'json-' to determine the redis key has not been found in the redis notification !\n");
+			goto err;
+		}
```
inf265: Yes right, it's not missing and it's not very nice but will be replaced when I removed the old multikey behaviour in a subsequent change (I am already working on it)

rfuchs:
So when trying this (start rtpengine, launch some calls, kill it and restart it), I run into a bunch of errors. Only one out of 20 calls could be restored, the others failed with
```
[1485882301.668752] DEBUG: Processing call ID 'json-YYyJZ5npL9VIfisrp5bPrP9u1fw5tLRRQg3U0uVCMizzcgaUTb' from Redis
[1485882301.668756] DEBUG: Processing call ID 'json-h7GZ8SbT5uhsZ6WPE6GQotiBWfbbqMweAtmY4lWzONs4TKjuT4' from Redis
[1485882301.668990] DEBUG: Could not parse json data !
[1485882301.668990] DEBUG: Could not parse json data !
...
[1485882301.669702] WARNING: Failed to restore call ID 'ysokNURgnpQ1sZQS36NFzpjr8bxnr4nOVrn7m7xucFt919AdwU' from Redis: failed to create sfds
[1485882301.669702] WARNING: Failed to restore call ID '5vIGFXe5T4HfBBCdxNGxqheMPtFAXo4pGkypUZlhIs0SZMGclC' from Redis: failed to create sfds
```

There seems to be a problem with storing binary data (encryption keys) into JSON format, perhaps related to unicode/utf8 encoding issues.

Sample JSON body of a failed call, straight out of redis:
```
{"json-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm":{"created":"1485882283","last_signal":"1485882283","tos":"0","deleted":"0","num_sfds":"8","num_streams":"4","num_medias":"2","num_tags":"2","num_maps":"2","ml_deleted":"0","created_from":"[::1]:56850","created_from_addr":"::1","redis_hosted_db":"0"},"sfd-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-0":{"pref_family":"IP6","localport":"30426","logical_intf":"default","local_intf_uid":"0","stream":"0","-crypto_suite":"AES_CM_128_HMAC_SHA1_80","-master_key":"2KAxLBIURybEs6XAhAxSSSQ17Xs008","-master_salt":"hAxSSSQ17Xs008","-unenc-srtp":"0","-unenc-srtcp":"0","-unauth-srtp":"0","last_index":"37369","ssrc":"4043817941"},"sfd-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-1":{"pref_family":"IP6","localport":"30427","logical_intf":"default","local_intf_uid":"0","stream":"1","-crypto_suite":"AES_CM_128_HMAC_SHA1_80","-master_key":"2KAxLBIURybEs6XAhAxSSSQ17Xs008","-master_salt":"hAxSSSQ17Xs008","-unenc-srtp":"0","-unenc-srtcp":"0","-unauth-srtp":"0","last_index":"0","ssrc":"0"},"sfd-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-2":{"pref_family":"IP6","localport":"30434","logical_intf":"default","local_intf_uid":"1","stream":"0"},"sfd-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-3":{"pref_family":"IP6","localport":"30435","logical_intf":"default","local_intf_uid":"1","stream":"1"},"sfd-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-4":{"pref_family":"IP6","localport":"30438","logical_intf":"default","local_intf_uid":"0","stream":"2"},"sfd-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-5":{"pref_family":"IP6","localport":"30439","logical_intf":"default","local_intf_uid":"0","stream":"3"},"sfd-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-6":{"pref_family":"IP6","localport":"30450","logical_intf":"default","local_intf_uid":"1","stream":"2"},"sfd-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-7":{"pref_family":"IP6","localport":"30451","logical_intf":"default","local_intf_uid":"1","stream":"3"},"stream-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-0":{"media":"0","sfd":"0","rtp_sink":"2","rtcp_sink":"4294967295","rtcp_sibling":"1","last_packet":"1485882298","ps_flags":"15794176","component":"1","endpoint":"[2607:fea8:a9df:fa04::90]:6390","advertised_endpoint":"[2607:fea8:a9df:fa04::90]:6390","stats-packets":"752","stats-bytes":"129344","stats-errors":"0","-crypto_suite":"AES_CM_128_HMAC_SHA1_80","-master_key":"ÞKm\u0005%\\ûIÉNëÌ\u0014¬Ýo\u0002ú\"\u0019CS\b\u0005/\bfãH","-master_salt":"o\u0002ú\"\u0019CS\b\u0005/\bfãH","-unenc-srtp":"0","-unenc-srtcp":"0","-unauth-srtp":"0","last_index":"56060","ssrc":"3221943417"},"stream-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-1":{"media":"0","sfd":"1","rtp_sink":"4294967295","rtcp_sink":"3","rtcp_sibling":"4294967295","last_packet":"1485882293","ps_flags":"15859712","component":"2","endpoint":"[2607:fea8:a9df:fa04::90]:6391","advertised_endpoint":"[2607:fea8:a9df:fa04::90]:6391","stats-packets":"3","stats-bytes":"360","stats-errors":"0","-crypto_suite":"AES_CM_128_HMAC_SHA1_80","-master_key":"ÞKm\u0005%\\ûIÉNëÌ\u0014¬Ýo\u0002ú\"\u0019CS\b\u0005/\bfãH","-master_salt":"o\u0002ú\"\u0019CS\b\u0005/\bfãH","-unenc-srtp":"0","-unenc-srtcp":"0","-unauth-srtp":"0","last_index":"4","ssrc":"0"},"stream-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-2":{"media":"1","sfd":"4","rtp_sink":"0","rtcp_sink":"4294967295","rtcp_sibling":"3","last_packet":"1485882298","ps_flags":"15794176","component":"1","endpoint":"[2607:fea8:a9df:fa04::90]:26528","advertised_endpoint":"[2607:fea8:a9df:fa04::90]:26528","stats-packets":"753","stats-bytes":"137046","stats-errors":"0"},"stream-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-3":{"media":"1","sfd":"5","rtp_sink":"4294967295","rtcp_sink":"1","rtcp_sibling":"4294967295","last_packet":"1485882298","ps_flags":"15859713","component":"2","endpoint":"[2607:fea8:a9df:fa04::90]:26529","advertised_endpoint":"[2607:fea8:a9df:fa04::90]:26529","stats-packets":"4","stats-bytes":"536","stats-errors":"0"},"stream_sfds-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-0":["0","2"],"stream_sfds-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-1":["1","3"],"stream_sfds-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-2":["4","6"],"stream_sfds-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-3":["5","7"],"tag-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-0":{"created":"1485882283","active":"1","deleted":"0","tag":"ZDfbPmo5Dmkt29T"},"tag-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-1":{"created":"1485882283","active":"0","deleted":"0","tag":"h4BJDlBQ6ycEhWK"},"other_tags-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-0":["1"],"medias-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-0":["1"],"other_tags-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-1":["0"],"medias-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-1":["0"],"media-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-0":{"tag":"1","index":"1","type":"audio","protocol":"UDP/TLS/RTP/SAVP","desired_family":"IP6","sdes_in_tag":"1","sdes_out_tag":"1","logical_intf":"default","media_flags":"2686988","sdes_in-crypto_suite":"AES_CM_128_HMAC_SHA1_80","sdes_in-master_key":"2KAxLBIURybEs6XAhAxSSSQ17Xs008","sdes_in-master_salt":"hAxSSSQ17Xs008","sdes_in-unenc-srtp":"0","sdes_in-unenc-srtcp":"0","sdes_in-unauth-srtp":"0","sdes_out-crypto_suite":"AES_CM_128_HMAC_SHA1_80","sdes_out-master_key":"ÞKm\u0005%\\ûIÉNëÌ\u0014¬Ýo\u0002ú\"\u0019CS\b\u0005/\bfãH","sdes_out-master_salt":"o\u0002ú\"\u0019CS\b\u0005/\bfãH","sdes_out-unenc-srtp":"0","sdes_out-unenc-srtcp":"0","sdes_out-unauth-srtp":"0"},"media-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-1":{"tag":"0","index":"1","type":"audio","protocol":"RTP/AVP","desired_family":"IP6","sdes_in_tag":"0","sdes_out_tag":"0","logical_intf":"default","media_flags":"196620"},"streams-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-0":["0","1"],"maps-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-0":["0"],"payload_types-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-0":["111/opus/48000/2","0/PCMU/8000/(null)","8/PCMA/8000/(null)"],"streams-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-1":["2","3"],"maps-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-1":["1"],"payload_types-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-1":["111/opus/48000/2","0/PCMU/8000/(null)","8/PCMA/8000/(null)"],"map-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-0":{"wildcard":"0","num_ports":"2","intf_preferred_family":"IP6","logical_intf":"default","endpoint":"[2607:fea8:a9df:fa04::90]:26528"},"map-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-1":{"wildcard":"0","num_ports":"2","intf_preferred_family":"IP6","logical_intf":"default","endpoint":"[2607:fea8:a9df:fa04::90]:6390"},"map_sfds-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-0":["loc-0","0","1","loc-1","2","3"],"map_sfds-ZBKcDPm6v1Ti06HmONMdiYYYf9I9IO7KqGDw1VWfNXlHBPfjlm-1":["loc-0","4","5","loc-1","6","7"]}
```

pkuzak:
Ow, that's not nice to hear. Indeed we didn't test any encrypted stuff. Regular (not encrypted) calls should work fine. We also kind of thought encryption is not yet fully supported to work with Redis, e.g. because of issue #303 . Nonetheless, a fix we can think of is to store the binary data _base64_ encoded. Would this be O.K. for you? 
It's not straightforward for us to construct a DTLS test case, do you maybe have a command line tool/scenario we can use for testing?

rfuchs:
DTLS is what's not supported, but SDES ought to work, and it looks like this is what this patch breaks.

Base64 for binary data should work, but I think there's a more fundamental issue here. As far as I know, JSON in general and JSON-Glib in particular expects all strings to be UTF8 encoded unicode strings, while none of rtpengine's internal strings are.

In practice, this _probably_ isn't an issue for most(?) strings, but there's no inherent guarantee that none of the strings contain any special characters. If a string is guaranteed to be plain ASCII only, it can be left as it is, but everything else ought to be escaped or guarded in some way and treated as bare binary data. The safest option would be to treat all strings as binary data, which would be fairly easy to achieve using a handful of wrapper functions.

Base64 for strings that are mostly or almost always (but can't be guaranteed to be) plain ASCII is obviously not a good choice. Perhaps using URI encoding or something similar would work better. My preference would be to use a single encoding mechanism for all strings and not get into trying to distinguish between what could and couldn't contain special characters.

rfuchs:
Oh, and for testing you can use the included (but poorly documented) `simulator-ng.pl` script. A sample command line from the repo's main directory is
```
perl -Iperl tests/simulator-ng.pl --local-ip=192.168.... --local-ipv6=2607:...  --num-calls=20 --runtime=60 --keep-going
```
There's a bunch of other options (in particular `--destination` might be of interest) listed in the head of the script file. Then while it's running, kill the daemon and restart it.

inf265:
Hi Richard,
thanks for the perl test line, that'll make things much easier for us. So I'll take also care about the UTF-8 stuff (along with my colleague Pawel).
Cheers
Frédéric

inf265:
Hi Richard,
I can see that i.e. the master key is generated with 30 bytes. With 'redis-multikey', you write only the first 16 to redis. I don't understand this. Could you please shorty explain. Thanks.

rfuchs:
The master key is 16 bytes and the master salt is 14 bytes. Makes 30 bytes together. In some cases (e.g. SDES SDP) they're communicated as a single 30 bytes long string.

There are some additional issues here, including a memory leak AFAICT. The URI encoding as it is now doesn't work fully correctly as it doesn't handle null bytes. I'm working on a comprehensive fix. Stay tuned.

inf265:
Thanks Richard. So I will wait for you fix if I understood you correctly.

rfuchs:
Check out the last commit in branch https://github.com/sipwise/rtpengine/tree/1and1-redis-onekey-concept-master

rfuchs:
https://github.com/sipwise/rtpengine/commit/f77726caa84777ffc5ef1357d80b3875d19dfa4b

inf265:
Ha great I thought before to make to make the de/en-code functions myself but I discarded. I will have a detailed look t'morrow thx a lot. And a nice evening.

rfuchs:
Well, that wasn't entirely by choice, but rather because there doesn't seem to be an existing library function that does the job :stuck_out_tongue: 

What I haven't tried is whether the Redis subscription still works. I haven't changed much, only made the pattern matching more precise, so it should be OK.

pkuzak:
I've tested the subscription/notification stuff and I can confirm that it is working (using f77726c).

inf265:
Thanks Richard.

PR #323, opened 2017-02-23
Redis onekey concept improvements

inf265:
This pull request:
* removes the multikey storing of data in the redis db
* removes the artifact of the redisReply used before when restoring a call (restore and notification)

Calls are now stored in redis under the key <callid> without prefix.


daemon/redis.c
```diff
@@ -812,7 +726,6 @@ static int json_get_hash(struct redis_hash *out, struct call* c,
 	out->ht = g_hash_table_new_full(g_str_hash, g_str_equal, free, freeReplyObject);
```
rfuchs: The hash table values aren't redis reply objects any more, so using `freeReplyObject` as value free function is incorrect.

Several points here.

The `redis_hash->rr` pointer used to contain the redis object containing the entire hash (in array form) for string storage. Replacing an entire hash/array with a single string makes no sense. So the `redis_hash` struct doesn't need to contain a string, but rather should contain only a `GHashTable`. I'll make separate comments on all relevant code lines.

For now I think the best approach is to `str_dup` the string values into the hash table (as you do it now) and use `free` as the value free function. (Since `json_reader_get_string_value_uri_enc` also mallocs its buffer, this can then be improved into a single malloc - I'll do this later on)

daemon/redis.c
```diff
@@ -924,15 +775,15 @@ static void json_destroy_list(struct redis_list *rl) {
...
 
 	r = g_hash_table_lookup(h->ht, k);
 	if (!r) {
 		out->s = NULL;
 		out->len = 0;
 		return -1;
 	}
-	out->s = r->str;
+	out->s = r->s;
 	out->len = r->len;
```
rfuchs: Using `*out = *r` is sufficient to copy all the string's fields in one go

daemon/redis.c
```diff
@@ -1734,37 +1386,21 @@ static int json_link_maps(struct redis *r, struct call *c, struct redis_list *ma
...
-	str s;
 	JsonReader *root_reader =0;
 	JsonParser *parser =0;
 
-	rr_jsonStr = redis_get(r, REDIS_REPLY_STRING, "GET json-" PB, STR(callid));
+	rr_jsonStr = redis_get(r, REDIS_REPLY_STRING, "GET " PB, STR(callid));
 	if (!rr_jsonStr) {
-		rlog(LOG_ERR, "Could not retrieve json data from redis for callid: " STR_FORMAT,
-				STR_FMT(callid));
+		rlog(LOG_ERR, "Could not retrieve json data from redis for key: %s", callid->s);
```
rfuchs: Please always use the `STR_FORMAT` and `STR_FMT()` macros when dealing with `str` objects. There's no inherent guarantee that the `->s` buffer is null terminated.

daemon/redis.c
```diff
@@ -2043,21 +1553,15 @@ static void restore_thread(void *call_p, void *ctx_p) {
 	redisReply *call = call_p;
 	struct redis *r;
 	str callid;
+	str_init(&callid,call->str);
```
rfuchs: use `str_init_len(&callid, call->str, call->len);` (as was removed below) to avoid an unnecessary `strlen()`

daemon/redis.h
```diff
@@ -62,7 +63,7 @@ struct redis {
 	int		no_redis_required;
 };
 struct redis_hash {
-	redisReply *rr;
+	str s;
```
rfuchs: So this can be removed entirely

daemon/redis.c
```diff
@@ -823,16 +736,11 @@ static int json_get_hash(struct redis_hash *out, struct call* c,
...
 			goto err3;
 		}
-		out->rr = createReplyObject(REDIS_REPLY_STRING);
-		if (!out->rr) {
-			rlog(LOG_ERROR, "Could not create redis reply object (out of memory?)");
-			goto err3;
-		}
-		out->rr->str = json_reader_get_string_value_uri_enc(c->root_reader, &out->rr->len);
-
+		str_init(&out->s,(char*)json_reader_get_string_value_uri_enc(c->root_reader,&out->s.len));
```
rfuchs: So this can be stored into a temporary `str` variable on the stack

inf265: changed

daemon/redis.c
inf265: changed

daemon/redis.c
inf265: changed

daemon/redis.c
inf265: changed

daemon/redis.h
inf265: removed

daemon/redis.c
inf265: Ok thanks for the comments. the "single malloc" I understood that you like to do it, if not just drop a quick message.

rfuchs:
Thanks guys. See master for slightly updated version.

PR #328, opened 2017-03-01
base64: flush base64 decoding, and skip base64 padding in crypto line

aalba6675:
Tweak base64:
- in case key length + salt length is not a multiple of 3, e.g., AES-192, AES-256
- skip base64 padding ("==" or "=")

### DIFF
--- a/daemon/sdp.c
+++ b/daemon/sdp.c
@@ -446,10 +446,9 @@ static int parse_attribute_crypto(struct sdp_attribute *output) {
 			(guchar *) c->key_salt_buf, &b64_state, &b64_save);
         // flush b64_state needed for AES-192: 36+2; AES-256: 45+1;
         if (enc_salt_key_len % 4) {
-                ret += g_base64_decode_step("==", 4 - b64_state,
+                ret += g_base64_decode_step("==", 4 - (enc_salt_key_len % 4),
                         (guchar *) c->key_salt_buf + ret, &b64_state, &b64_save);
         }
-        assert( !b64_state );
 	err = "invalid base64 encoding";
 	if (ret != salt_key_len)
 		goto error;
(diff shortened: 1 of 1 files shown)


daemon/sdp.c
```diff
@@ -444,6 +444,12 @@ static int parse_attribute_crypto(struct sdp_attribute *output) {
 	str_shift(&c->key_base64_str, 7);
 	ret = g_base64_decode_step(c->key_base64_str.s, enc_salt_key_len,
 			(guchar *) c->key_salt_buf, &b64_state, &b64_save);
+        // flush b64_state needed for AES-192: 36+2; AES-256: 45+1;
+        if (enc_salt_key_len % 4) {
+                ret += g_base64_decode_step("==", 4 - b64_state,
```
rfuchs: I would prefer not to use `b64_state` directly since the glib docs suggest it to be an opaque value and the `g_base64_decode_step` function may actually put a negative number there. It should be possible to use `enc_salt_key_len` to determine how much padding is needed (e.g. something like `4 - (enc_salt_key_len % 4)`)

aalba6675:
Good catch — I will fix this; bad bad bad UAs send padding :-(
From pjproject v2.6 pjsua
a=crypto:1 AES_CM_256_HMAC_SHA1_80 inline:c2fQwFT9GEtwYlyUA9q4dRW/nAu7q0lSqh4sqzfjiezQQA9DtVmR4d+iJvVWYQ==

From FreeSWITCH 1.6.15
a=crypto:1 AES_CM_256_HMAC_SHA1_80 inline:NazjxRMjncLYVcuMYzIK80qsiDFswLT70gDEYn1qW+QtqETm67c07hXD72WI1O==

aalba6675:
Fixed commit per @rfuchs comment. Did not touch encoding part.

rfuchs:
Hm you may be right. I misread the RFC. Referring to RFC 4568 section 6.1 https://tools.ietf.org/html/rfc4568#section-6.1 it states:

When base64 decoding the key and
   salt, padding characters (i.e., one or two "=" at the end of the
   base64-encoded data) are discarded 

This talks about decoding, not encoding as I originally thought. Also the attribute grammar from section 9.2 includes `=` as a valid base64 character, so I think leaving the padding in place is the correct thing to do.

PR #332, opened 2017-03-18
Initial support for AES-192, AES-256 in userspace

aalba6675:
This is an initial attempt at AES-192 AES-256 support for comments. It works in userspace / SDES only.
Remaining issues:
* no DTLS (the equivalent DTLS suites seem to have died in IETF drafts)
* kernel support; will need to pass key length some how to kernel interface; open to suggestions here


rfuchs:
Nice, thank you. Couple of comments (of course :smile:)

I would prefer to use separate function pointers for the different key lengths instead of using the same one and doing a case distinction based on key length. Something like changing `aes_cm_session_key_init` to `aes_cm_session_key_init(struct crypto_context *c, EVP_CIPHER *ecb_cipher)` and then have something like
```
static int aes_cm_session_key_init_128(struct crypto_context *c) {
  return aes_cm_session_key_init(c, EVP_aes_128_ecb())
}
```
and so on, and use these in `crypto_suite` definitions. Does that make sense? A similar approach should be possible with `aes_ctr_128_no_ctx`. Another solution could be to include the `EVP_CIPHER *` pointer in the `crypto_suite` definitions directly.

Does the `aes_ctr_128()` function also need to be adapted?

As for kernel support, it can either follow the same mantra and extend `re_ciphers` to include separate entries for different key lengths (rename `REC_AES_CM` to `REC_AES_CM_128` and add `REC_AES_CM_192` etc) and have separate encoding functions, or extend `struct rtpengine_srtp` to include the key length. I haven't checked what would make more sense. If there's no case distinction necessary, then extending the struct with the key length should be sufficient.

I can make these changes if you'd prefer.

aalba6675:
Refactor:
1.  Add explicit aes_cm_session_key_init_NNN functions
1. Rename aes_ctr_128() to aes_ctr(), aes_ctr_128_no_ctx to aes_ctr_no_ctx()
1. aes_ctr (was aes_ctr_128) seems to be ok (i.e., key length unaware), as it just uses the passed-in cipher context
1. crypto_gen_session_key() is in the public interface: I did not change that, and am extracting key length to pass to prf_n(). Any qualms about this one? 
1. I'll next look at the kernel interfaces...

aalba6675:
Initial kernel support:

1. Pass master key len and session key len to kernel structures; let crypto_cipher_setkey() handle AES-128, AES-192, AES-256 key lengths

rfuchs:
Hi, I've made some additional changes to this, see code in branch https://github.com/sipwise/rtpengine/tree/rfuchs/aes
I don't have a test case for AES-192/256, so if possible, please confirm that this is still working correctly, thanks.

aalba6675:
I couldn't reach the branch https://github.com/sipwise/rtpengine/tree/rfuchs/aes
Am getting 404 errors

rfuchs:
Durr, it was deleted by our gerrit by accident. It's there now.

aalba6675:
Sure — let me test this; meanwhile, I have pushed a commit that fixes a stupid indexing error in DBG: 21->31

rchan1611:
Added AES static test vectors in tests/. HOWTO:
561. build daemon/
2. cd tests/
3. make -f Makefile.aes
4. ./t_crypt

Test vectors taken from libsrtp, added AES-192, AES-256 computations.
Test are using SRTCP_R_LENGTH == 6

aalba6675:
Added AES static test vectors in tests/. HOWTO:
1. build daemon/
2. cd tests/
3. make -f Makefile.aes
4. ./t_crypt

Test vectors taken from libsrtp, added AES-192, AES-256 computations.
Test are using SRTCP_R_LENGTH == 6

Expected behaviour
```shell
$ ./t_crypt
SRTP AES-CM-128 RTP encrypt: PASS
SRTP AES-CM-128 RTP decrypt: PASS
SRTP AES-CM-128 RTCP encrypt: PASS
Added AES static test vectors in tests/. HOWTO:
1. build daemon/
2. cd tests/
3. make -f Makefile.aes
4. ./t_crypt

Test vectors taken from libsrtp, added AES-192, AES-256 computations.
Test are using SRTCP_R_LENGTH == 6

Expected behaviour
```shell
$ ./t_crypt
SRTP AES-CM-128 RTP encrypt: PASS
SRTP AES-CM-128 RTP decrypt: PASS
...
Added AES static test vectors in tests/. HOWTO:
1. build daemon/
2. cd tests/
3. make -f Makefile.aes
4. ./t_crypt

Test vectors taken from libsrtp, added AES-192, AES-256 computations.
Test are using SRTCP_R_LENGTH == 6

Expected behaviour
```shell
$ ./t_crypt
SRTP AES-CM-128 RTP encrypt: PASS
SRTP AES-CM-128 RTP decrypt: PASS
SRTP AES-CM-128 RTCP encrypt: PASS
...
extra AES-CM-256 RTP encrypt: PASS
extra AES-CM-256 RTP decrypt: PASS

```

PR #335, opened 2017-03-21
Remove jsonreader from call

pkuzak:
Resolving technical debt introduced in #316 (_root_reader_ being part of call object).

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -1841,6 +1841,7 @@ void call_destroy(struct call *c) {
 	/* at this point, no more packet streams can be added */
 
 	if (IS_OWN_CALL(c)) {
+		ilog(LOG_INFO, "Final packet stats:");
 
 		for (l = c->monologues.head; l; l = l->next) {
 			ml = l->data;
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -38,7 +38,8 @@
 #include "log_funcs.h"
 #include "recording.h"
 #include "rtplib.h"
-
+#include "cdr.h"
+#include "statistics.h"
 
 
 /* also serves as array index for callstream->peers[] */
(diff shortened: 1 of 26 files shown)


rfuchs:
This seems to include the commits from #334 as well. Do you want to close the other pull request?

inf265:
Yes but anyway I don't have the permission I guess ?

inf265:
Ahh I found

daemon/call.c
```diff
@@ -113,37 +114,8 @@ const struct transport_protocol transport_protocols[] = {
...
-};
-static const char *const __opmode_texts[] = {
-	[OP_OFFER] = "offer",
-	[OP_ANSWER] = "answer",
-};
-
-static const char * get_term_reason_text(enum termination_reason t) {
-	return get_enum_array_text(__term_reason_texts, t, "UNKNOWN");
-}
-const char * get_tag_type_text(enum tag_type t) {
```
rfuchs: Can you also please move the relevant header definitions from `call.h` to `cdr.h`

inf265: Yes but to be honest I do not very like the parameter handover deep down to every function but I think it's better than before. We'll see if there's a more better way in the future.

inf265:
Yes but to be honest I do not very like the parameter handover deep down to every function but I think it's better than before. We'll see if there's a more better way in the future.

daemon/call.c
inf265: I'll have a look tmorrow and sure I'll move :-D

rfuchs:
Yes I agree, that's why I thought using a global TLS variable would avoid the extra parameter everywhere. At global level, you would declare `static __thread JsonReader *root_reader`, then set this variable at the beginning of the restore process and clear it at the end. Because it's TLS, it's safe to use in all participating functions.

Another option would be to use a helper struct that contains the `call` struct pointer and the `JsonReader` pointer and pass that struct to every function instead of the `call` pointer. You'd just replace `struct call *c` with `struct foobar_helper *x` everywhere and use `x->c` instead of `c` to access the `call`.

Either way, it's up to you. Just some ideas. I'm OK with merging it as it is.

inf265:
Ok if it's ok we leave it as it is. And thanks for explaining the other options.

daemon/call.c
inf265: moved

PR #346, opened 2017-04-05
make cli_incoming accept multiple connections

### LINKED ISSUE #280: rtpengine-ctl stops returning / printing any data
We've started running rtpengine with a bit more load (only 50 or sessions) and one of our relays has stopped printing any results when using rtpengine-ctl.

```
$rtpengine-ctl list numsessions
$

```

The logs show the command is being received ...

```
2016-10-11T16:45:50.441932+00:00 media-relay2-0 rtpengine[8263]: INFO: New cli connection from 127.0.0.1:36436
2016-10-11T16:45:50.441941+00:00 media-relay2-0 rtpengine[8263]: INFO: Got CLI command:list numsessions
```

It just doesn't print any thing.  I've tried various commands.  I had been running various list numsessions commands throughout the morning and it was all working fine.  The box is still processing traffic, and relaying rtp without issue.  Is there any way to ... "reset" the cli without having to restart rtpengine, or do you have any other suggestions?

Thank you.

claudiupb:
fixes #280 
The original problem in this issue is generated because multiple incoming connections on the listening socket can generate a single epoll event. Since the events have EPOLLET flag this means that if only one connection is treated (as with the code before) the other ones will never be processed, because no other event will be generated for them. If this happens some incoming connections will never be treated and the client will never get a response (this is why the rtpengine-ctl client never prints and remains hanged).  
Modifications to the cli_incoming function will allow to treat all connections, since now accept is called until it returns EAGAIN || EWOULDBLOCK.

### DIFF
--- a/daemon/cli.c
+++ b/daemon/cli.c
@@ -840,7 +840,6 @@ next:
    nfd = accept(fd, (struct sockaddr *) &sin, &sinl);
    if (nfd == -1) {
        if (errno == EAGAIN || errno == EWOULDBLOCK) {
-           ilog(LOG_INFO, "Could currently not accept CLI commands. Reason:%s", strerror(errno));
            goto cleanup2;
        }
        ilog(LOG_INFO, "Accept error:%s", strerror(errno));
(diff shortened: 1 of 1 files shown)


rfuchs:
Good catch, thanks. The log messages from line 843 should also be removed since this is now an expected condition.

claudiupb:
Yes, I removed the message.

PR #353, opened 2017-04-24
More control over iptables

pkuzak:
Relates to #205.

### DIFF
--- a/debian/ngcp-rtpengine-daemon.default
+++ b/debian/ngcp-rtpengine-daemon.default
@@ -2,6 +2,7 @@ RUN_RTPENGINE=no
 CONFIG_FILE=/etc/rtpengine/rtpengine.conf
 # CONFIG_SECTION=rtpengine
 PIDFILE=/var/run/ngcp-rtpengine-daemon.pid
+MANAGE_IPTABLES=yes
 TABLE=0
 #SET_USER=root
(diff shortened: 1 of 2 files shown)


rfuchs:
Hm, I'm not sure about this one. It seems to lose the meaning of the `CREATE_IPTABLES_CHAIN` switch. Some people might be using that option. Can you elaborate on your rationale behind this change?

Also, this looks like a good opportunity to get rid of all the old references to `mediaproxy`.

pkuzak:
Sure. this change was motivated by our operational/administration team (as was #205, but #205 didn't solve all their requirements). They are managing all _iptables_ rules separately. Also, they take care of loading and unloading Kernel modules. So they asked me whether I can extend the _init_ script accordingly. That's why I came up with this pull request. 
I renamed (or deleted) the old _CREATE_IPTABLES_CHAIN_ switch because that one only controlled the creation of _iptables_ rules. The deletion was always done. I thought the name _MANAGE_IPTABLES_ is more suitable.

rfuchs:
One problem that I can see is that the `rmmod` must come after the `del $TABLE`. I'll fix it up and commit it.

rfuchs:
Although it seems that you've moved the `rmmod` below the `del $TABLE` intentionally. Any particular reason for that?

pkuzak:
Oh...O.K.. Honestly, looking at the code I don't know why I moved this part down in the code. Thank you for taking care of this.

PR #402, opened 2017-10-10
set TOS for redis streams

claudiupb:
In case kernel forwarding is disabled, TOS should be set for streams that are created from redis db.


daemon/redis.c
```diff
@@ -1067,6 +1067,7 @@ static int redis_sfds(struct call *c, struct redis_list *sfds) {
 		sock = g_queue_pop_head(&q);
 		if (!sock)
 			goto err;
+		set_tos(sock, c->tos);
 		sfd = stream_fd_new(sock, c, loc);
 		// XXX tos
```
rfuchs: Should also remove this comment as this makes it obsolete :)

PR #403, opened 2017-10-10
set TOS for redis streams

claudiupb:
In case kernel forwarding is disabled, TOS should be set for streams that are created from redis db.


claudiupb:
I made a new Pull Request since a forced push was not visible in the old one.

PR #411, opened 2017-10-25
Add recording forwarding

claudiupb:
These commits add the option for recording daemon to forward the captured traffic to a UNIX domain socket instead of encoding to wav/mp3 and saving an audio file.  We chose a UNIX domain socket because we can use the SOCK_SEQPACKET type which allows to make have a connection based communication, but uses datagrams instead of streams, so it will be easier to detect each individual RTP package. 
This way for each call a new connection is made and a simple program that listens on than UNIX domain socket can process each RTP packet from each stream, as needed. 

The first received message contains the metadata info in clear text format, so control information can easily be passed from rtpengine (or from kamailio by using start_recording() or rtpengine_manage()).

### DIFF
--- a/daemon/call_interfaces.c
+++ b/daemon/call_interfaces.c
@@ -747,9 +747,9 @@ static const char *call_offer_answer_ng(bencode_item_t *input, struct callmaster
 	chopper = sdp_chopper_new(&sdp);
 	bencode_buffer_destroy_add(output->buffer, (free_func_t) sdp_chopper_destroy, chopper);
 
-	detect_setup_recording(call, &flags.record_call_str);
+	detect_setup_recording(call, &flags.record_call_str, &flags.metadata);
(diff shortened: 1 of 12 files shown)


rfuchs:
May I ask about the motivation behind this feature? I'm wondering because the RTP packets are already available for reading locally through the /proc interface, so having the recording-daemon read them locally just to re-deliver them elsewhere locally seems like an unnecessary step.

claudiupb:
The recording-daemon also parses the .meta files from the spool directory, and has the stream structures for each call in place. If another application will want tho read the /proc interface, it will also need to monitor the file-system to read and parse these files so it knows the number of streams for each call and where to read them. All this is already implemented in the recording daemon, so by using this feature any application that would simply want to read the RTP packets organized by calls will have a reduced complexity.

PR #416, opened 2017-11-06
don't use g_hash_table_foreach for call_timer_iterator

claudiupb:
this keeps the callmaster->hashlock for a long time so copy the list of calls instead and update call refs


rfuchs:
Nice catch, thank you

PR #420, opened 2017-11-17
cli params command has beend created for initial, current parameters.

hdikme:
Using the below commands, it has been enabled to screen initial & present values of rtpengine.conf parameters. Additionally, it is possible to now check if any parameter has been modified.
"rtpengine-ctl params start"
"rtpengine-ctl params current"
"rtpengine-ctl params diff"

### DIFF
--- a/daemon/call.h
+++ b/daemon/call.h
@@ -21,6 +21,7 @@
 #include "statistics.h"
 
 #define UNDEFINED ((unsigned int) -1)
+#define MAX_SIZE 255
 
 enum termination_reason {
 	UNKNOWN=0,
@@ -386,6 +387,36 @@ struct call {
 
 	struct recording 	*recording;
 };
+struct callmaster_cfg_addition {
+	int table;
+	int no_fallback;
+	int port_min;
+	int port_max;
+	int no_redis_required;
+	int num_threads;
+	int sip_source;
+	int dtls_passive;
+	int homer_id;
+
+	char listenps[MAX_SIZE];
+	char listenudps[MAX_SIZE];
+	char listenngs[MAX_SIZE];
+	char listencli[MAX_SIZE];
+	char graphite[MAX_SIZE];
+	char graphite_prefix[MAX_SIZE];
+	char redis[MAX_SIZE];
+	char redis_write[MAX_SIZE];
+	char log_facility_rtcp[MAX_SIZE];
+	char log_facility_cdr[MAX_SIZE];
+	char homer[MAX_SIZE];
+	char homer_protocol[MAX_SIZE];
+	char rec_dir[MAX_SIZE];
+	char rec_method[MAX_SIZE];
+	char rec_format[MAX_SIZE];
+
+	GQueue interfaces ;
+};
+
 
 struct callmaster_config {
 	/* everything below protected by config_lock */
@@ -410,6 +441,7 @@ struct callmaster_config {
 	int			graphite_interval;
 
 	int			redis_num_threads;
+	struct callmaster_cfg_addition cfg_addition;
 };
 
 struct callmaster {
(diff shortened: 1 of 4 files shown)


rfuchs:
Wow, this is a very large commit for what it's trying to achieve.

Only a small number of config parameters are currently mutable during runtime. Can you explain the reason for having this mechanism for parameters that cannot be changed?

hdikme:
Hello Richard,

While coding the "cli_incoming_params_diff" function, i have also thought that there is a limited number of parameters that can be changed at the run time but then i also thought those parameters could be increased by the upstream in the course of the time. In order not to be affected by any future modifications, the function has been implemented this way. If you think there is gonna be no more addition in terms of the parameters that could be changed at the run time, then i will reorganize the  "cli_incoming_params_diff" function. Thanks for your time.

rfuchs:
Fair enough. I have no problem with taking possible future expansions into account. However, the amount of code duplication between callmaster_config, callmaster_cfg_addition, and rtpengine_config_params, and related boilerplate code is far from ideal. A more unified and consolidated approach would be far better.

Why not combine everything into a single struct, and keep two global instances thereof (running config and startup config)? Then callmaster_config can be moved out of the callmaster struct, and the `mc` special handling can then be eliminated. (In fact, the entire callmaster struct should be thrown out the window and converted to global variables instead. But that's way beyond the scope of this change.)

All the strlen/strcpy stuff can be avoided by using g_strlcpy, or strdup (especially strdupfree from aux.h).

There's no reason the GOptionEntry list cannot reference struct members directly.

Is this something you'd be willing to work on? If not, I can do the preparation work and cleanup/removal of old code that serves no purpose myself, and then you can add your new functions on top of that.

hdikme:
As far as I understand, I will remove the callmaster_config struct from the callmaster struct and then add some variables (all additional variables from the callmaster_cfg_addition struct) into callmaster_config. Following that, I will create
2 instances of callmaster_config one for current the other for the initial config parameters. And that means “struct callmaster_config mc” at the beginning of function“static void create_everything(struct main_context *ctx)” will be moved outside the function and defined as a global struct. At this point I m wondering: 
1)callmaster config contains fields like config_lock, *redis, *redis_write that are not populated at startup. What happens with such fields? Should these be defaulted to NULL in the initial config?

2)What do you mean exactly by saying “Then callmaster_config can be moved out of the callmaster struct, and the mc special handling can then be eliminated.“? Because I thought that I should define mc outside the function as a global struct and the rest would be left the same but the word “eliminating” has confused me a bit.

3)Any yes I’m willing to implement this my own but what do you mean by doing the preparation work (removal of old code)?

rfuchs:
config_lock is used to synchronize access to certain fields of the (running) config. If a startup-config is read-only, it does not need a config_lock.

*redis and *redis_write are more handles than config options. There's a few config options in there (endpoint, db) and those should go into a config struct and be populated accordingly, but the redis handles themselves should be kept separately.

The purpose of the 'callmaster_config mc' is/was to populate the config struct before the callmaster itself is created, and then copy the fields from there into the callmaster's config. If the config is moved out of the callmaster and made global, this extra step becomes redundant and the global config struct can be populated directly.

And what I meant is that I can do all of the changes I just described myself if you'd prefer. I would aim for a clean(er) config handling, which would lend itself to keeping track changes between a startup config and the runtime config. It's up to you.

rfuchs:
FYI, I've created an internal ticket to do the config cleanup I was talking about. Stay tuned.

rfuchs:
Check out the code in https://github.com/sipwise/rtpengine/tree/rfuchs/config-rework

This gets rid of the callmaster struct, moves everything holding state information global, and moves all config options into a globally held `rtpe_config` struct.

Unparsed config options (such as the raw `listen-...` strings) are held only in local variables right now, but those could be moved into the global config struct as well if necessary.

The only ones left are the options from auxlib.c (such as the `log-level`) - I will gather them up and put them into the config struct as well.

hdikme:
Thank you for the updating, i m already working on the code, fyi.

hdikme:
Hi richard,
I have created a new pull-request based-on the new structure.
https://github.com/sipwise/rtpengine/pull/452

Issue #275, opened 2016-09-15
RTPEngine Re-Invite issue


mvenkatesh431:
Hi all, 
   I am using RTPEngine and OpenSIPS. It working great. But it is not working properly on Re-Invite cases.
Here is my scenario : 
A initiates a Video call to B.
B accepts only audio call without video. Now both are connected. B can see A's Video and A can hear B's audio.
Now B sends Re Invite with Video option.  A accept the call.  Now in this case A can see B's video but B can't see A's video. It is completly frozen. But after sometime like 90 sec or more video started transmitting properly. 

I am using RTPEngine and OpenSIPS and SIP.js as client side library.
RTPEngine Version : git-master-cdcf2ff
OpenSIPS version :  opensips 2.3.0-dev (x86_64/linux)

Here is the OpenSIPS script : 

if (has_totag()) {
     # sequential requests within a dialog should
     # take the path determined by record-routing
     if (loose_route()) {
             if (is_method("INVITE")) {
                       fix_nated_contact();
                       t_on_branch("handle_reinvite");
                       # even if in most of the cases is useless, do RR for
                       # re-INVITEs alos, as some buggy clients do change route set
                       # during the dialog.
                       xlog(" Re Invite is called ");
                       record_route();
             }
    }
}

if (is_method("INVITE"))
{
          t_on_branch("handle_nat");
          t_on_reply("handle_nat");
}

branch_route[handle_nat] {
        if (!is_method("INVITE") || !has_body("application/sdp"))
                return;
        $var(rtpengine_flags) = "ICE=force DTLS=passive";
     rtpengine_offer("$var(rtpengine_flags)");
}

branch_route[handle_reinvite] {
        if (!is_method("INVITE") || !has_body("application/sdp"))
                return;
        t_on_reply("handle_reinvite");
        $var(rtpengine_flags) = "ICE=force DTLS=passive";
        rtpengine_offer("$var(rtpengine_flags)");
}

onreply_route[handle_nat] {
        fix_nated_contact();
        if (!has_body("application/sdp"))
                return;
        fix_nated_sdp("3");
        $var(rtpengine_flags) = "ICE=force DTLS=passive";
    rtpengine_offer("$var(rtpengine_flags)");
}

onreply_route[handle_reinvite] {
        if (!has_body("application/sdp"))
                return;
        fix_nated_sdp("3");
        $var(rtpengine_flags) = "ICE=force DTLS=passive";
[...]

Issue #393, opened 2017-09-13
UnHold problem


falconsapsan:
JsSIP v3.0.15 * JsSIP v3.0.15
rtpengine version: 5.5.0.0+0~mr5.5.0.0

JsSIP -> Opensips (rtpengine) -> Freeswitch

# For Chrome, Opera
user A --> calling -->user B  (all OK)
user A --> hold  --> user B (all OK, user B hear hold music)
user A --> unhold -->user B  (user B does not hear user A)

# For Firefox
user A --> calling -->user B  (all OK)
user A --> hold  --> user B (all OK, user B hear hold music)
user A --> unhold -->user B  (user A does not hear user B)

This is  bug?

rfuchs:
You'd have to a post a full debug log (log level 7) to find out

egreenmachine:
I've tested renegotiation with the [SIP.js renegotiation branch](https://github.com/onsip/SIP.js/pull/426) and rtpengine. Rtpengine works flawlessly in the cases outlined above and more. I would be surprised if JsSIP were doing something different enough that would be causing rtpengine to be the problem. Specifically because the SDP is generated from the browser itself, not the library.

IvanKalashnikov:
Hello, get same problem, but use kamailio + freeswitch + rtpengine + jssip
Rtpengine Version: 7.4.0.0+0~mr7.4.0.0 git-master-285773f
Attach log with loglevel 7
[unhold.log](https://github.com/sipwise/rtpengine/files/3325742/unhold.log)

rfuchs:
@IvanKalashnikov the log doesn't show anything that would explain loss of audio. Please check on the sides of the RTP clients.

rfuchs:
Cleaning up old tickets

PR #432, opened 2017-12-15
Replace transport protocol only if protocol was specified in flags

arkadiam:
Setup: Polycom phone with SDES (optional) -- Kamailio + Rtpengine -- Asterisk PBX

On initial INVITE Polycom sends SDP with:
m=audio 2278 RTP/SAVP
m=audio 2278 RTP/AVP
m=video 2280 RTP/SAVP
m=video 2280 RTP/AVP

Kmailio calls rtpengine_manage() with flags:
"direction=pub direction=priv replace-origin replace-session-connection ICE=remove SDES-off"

INVITE passed to Asterisk with the same media options (different port numbers). Asterisk is configured only to accept audio RTP. Call is answered and media goes like:
Polycom -- SRTP -- rtpengine -- RTP -- Asterisk.

Now Polycom puts call on HOLD. That produce INVITE with SDP like:
m=audio 2278 RTP/SAVP
m=audio 2278 RTP/AVP
m=video 0 RTP/SAVP
m=video 0 RTP/AVP

At this point, by some reason, Rtpengine replaces transport protocol with RTP/AVP and Asterisk presented with such media options:
m=audio xxx RTP/AVP
m=audio xxx RTP/AVP
m=video 0 RTP/AVP
m=video 0 RTP/AVP
Multiple RTP/AVP for audio confuses the Asterisk and it rejects the INVITE with 488.

Patch in this pull request prevents protocol replacement if it wasn't set in "flags". 
It allows RE-INVITE to pass and present Asterisk valid media options:
m=audio xxx RTP/SAVP
m=audio xxx RTP/AVP
m=video 0 RTP/SAVP
m=video 0 RTP/AVP

This may potentially fix issues #275 and #393

### DIFF
--- a/daemon/sdp.c
+++ b/daemon/sdp.c
@@ -1952,12 +1952,14 @@ int sdp_replace(struct sdp_chopper *chop, GQueue *sessions, struct call_monologu
 			ps = j->data;
 
 			if (!flags->ice_force_relay) {
-			        if (replace_media_port(chop, sdp_media, ps))
-				        goto error;
-			        if (replace_consecutive_port_count(chop, sdp_media, ps, j))
-				        goto error;
-				if (replace_transport_protocol(chop, sdp_media, call_media))
-				        goto error;
+				if (replace_media_port(chop, sdp_media, ps))
+					goto error;
+				if (replace_consecutive_port_count(chop, sdp_media, ps, j))
+					goto error;
+				if (flags->transport_protocol) {
+					if (replace_transport_protocol(chop, sdp_media, call_media))
+						goto error;
+				}
 
 				if (sdp_media->connection.parsed) {
 				        if (replace_network_address(chop, &sdp_media->connection.address, ps,


rfuchs:
Can you post a concrete example of a call going wrong? Debug log containing the SDP bodies would be great.

I don't believe this is a the correct patch to fix this. Rtpengine is supposed to remember the protocols used by clients that it talked to before, and is supposed to insert the correct transport protocol without being instructed to (or even when not instructed to do anything).

arkadiam:
Thanks for looking at it. 
Please find the logs of a test call attached. 
It follows calls scenario I've originally described: Polycom phone places a call, call is answered, then Polycom phone puts call on Hold, then it Resumes the call.

The issue corresponds to the log line 53 where Rtpengine translates one of "m=audio" to RTP/AVP.

[rtpengine-20171218.log.txt](https://github.com/sipwise/rtpengine/files/1569271/rtpengine-20171218.log.txt)

rfuchs:
I believe the problems stems from the way Asterisk responds to the multi-stream SDP.

Polycom offers 4 media streams: audio SRTP, audio RTP, video SRTP, video RTP.
Asterisk then answers with only 2 media streams: audio RTP, video RTP (disabled)

My understanding of RFC 3264 (section 6) is that media descriptions in the SDP (m= lines) in the answer must match up with media descriptions given in the offer. So if the offer contains 4 media streams, the answer must also contain 4 media streams. Asterisk doesn't do this.

The result is that rtpengine matches up the first media stream from Polycom with the first media stream from Asterisk, thus it thinks that Polycom wants to speak SRTP there while Asterisk wants to speak RTP. It remembers this and in the subsequent re-offer substitutes the appropriate transport protocol.

A correct response from Asterisk (IMO and from my understanding of the RFC) would be to generate an SDP answer with 4 media streams, with the same types and protocols as the original offer, and all of them except the second one (audio RTP) disabled (zero port).

arkadiam:
Asterisk version we have to work with, doesn't support SRTP. Unsupported media are being ignored without producing zero-port m-line records. I agree with your RFC3264 interpretation.

I can try to modify Asterisk code to return 4 media streams like:
m=audio 0 RTP/SAVP
m=audio 2278 RTP/AVP
m=video 0 RTP/SAVP
m=video 0 RTP/AVP

BUT I'm not sure if in this case the media path will be:
    Polycom -- SRTP -- rtpengine -- RTP -- Asterisk.
I think it might be turned into:
    Polycom -- RTP -- rtpengine -- RTP -- Asterisk. 
which is not our intention. We would like to keep the first call leg secure.

I don't see how rtpengine_manage() can be called with any other flags to achieve SRTP-RTP conversion. Specifying transport protocol: "RTP/AVP"  will turn initial  SDP into:
m=audio 2278 RTP/AVP
m=audio 2278 RTP/AVP
m=video 0 RTP/AVP
m=video 0 RTP/AVP
which will be rejected by Asterisk due to duplicate values.

rfuchs:
No, rtpengine will remember that Polycom speaks SRTP in that media stream and will translate accordingly. No extra flags are needed.

rfuchs:
Ah no, actually you're right. It would be RTP both ways. I suppose in the `offer`, you have to specify `RTP/AVP`. (This sort of alternative protocol selection isn't supported by rtpengine. I'm not sure if there's an RFC for it, but I guess it could be implemented anyway.)

rfuchs:
A possible workaround for the Asterisk side of this is to make rtpengine recognize this situation (m= lines missing) and attempt to match up the lines as Asterisk thinks they should match up (possibly by internally filling up the missing media streams with dummy entries). See `__get_media` and its invocation in `monologue_offer_answer`

rfuchs:
#435 is related

rfuchs:
stale

PR #445, opened 2018-01-18
Add timeouts to redis commands, and possiblity to temporarily disable redis

claudiupb:
When having issues with the redis server the offer/answer/delete times will be affected.
With the new parameters a user can configure a timeout for redis commands and connection, and also to temporarily disable redis communication after a number of consecutive errors


README.md
```diff
@@ -462,6 +463,13 @@ The options are described in more detail below.
 	If redis does not reply	within the specified timeout the command will fail. The default value is 0, meaning that the commands
 	will have no timeout
 
+*  --redis-connect-timeout
+	This parameter sets the timeout value, in milliseconds, when connecting to a redis server. If the connection cannot be made
+	within the specified timeout the connection will fail. Note that in case of failure, when reconnecting to redis, a PING command
+	is issued before attempting to connect so the `--redis-cmd-timeout` value will also be added to the total waiting time.
+	This is useful if using `--redis-allowed-errors', when attempting to estimate the total lost time in case of redis failures.
```
rfuchs: Incorrect type of quote used here ( ' instead of ` )

rfuchs:
Can you explain a bit about the symptoms you're seeing? Would it be beneficial to run the redis commands asynchronously in a separate thread?

claudiupb:
The problem can appear if there is a slow connection to redis, and is some situations the offer/answer/delete processing time become too big and kamailio will disable the rtpengine node since it will not respond in time. I think it will be beneficial to use a separate thread for redis commands, but for using the asynchronous commands will require a rewrite of the redis functions to use the hiredis asynchronous API.

rfuchs:
Use of the asynchronous API might not be strictly necessary. Simply appending the Redis commands to a queue and then having a dedicated thread read from the queue and issue to commands to Redis synchronously would probably also work. Not sure which would be more effort to implement. Either way, just an idea.

PR #446, opened 2018-01-19
Add statistics for requests per second

claudiupb:
This commit adds statistics for the number of offer/answer/delete requests per second. For each second the number of requests is counted and then a min/max is kept and an average is computed. The statistics are also sent to the graphite server.


daemon/call.c
```diff
@@ -503,6 +522,19 @@ static void call_timer(void *ptr) {
 	atomic64_set(&rtpe_stats.packets, atomic64_get_na(&tmpstats.packets));
 	atomic64_set(&rtpe_stats.errors, atomic64_get_na(&tmpstats.errors));
 
+	/* update statistics regarding requests per second */
+	offers = atomic64_get(&rtpe_statsps.offers);
+	atomic64_set(&rtpe_statsps.offers, 0);
```
rfuchs: I suggest using `atomic64_get_set` instead of one `get` and one `set` here, as otherwise there's a chance of losing some counts. (Applies to all 3 occurrences)

daemon/cli.c
```diff
@@ -199,6 +200,9 @@ static void cli_incoming_list_totals(str *instr, struct streambuf *replybuffer)
 	offer_iv = rtpe_totalstats_lastinterval.offer;
 	answer_iv = rtpe_totalstats_lastinterval.answer;
 	delete_iv = rtpe_totalstats_lastinterval.delete;
+	offers_ps = rtpe_totalstats_lastinterval.offers_ps;
+	answers_ps = rtpe_totalstats_lastinterval.answers_ps;
+	deletes_ps = rtpe_totalstats_lastinterval.deletes_ps;
```
rfuchs: These aren't thread-safe, as 64-bit types aren't intrinsically atomic on all platforms, and these structs aren't protected by `rtpe_totalstats_lastinterval_lock`. Either lock each struct's mutex before copying the struct, or copy the values individually using atomic64 operations.

daemon/cli.c
```diff
@@ -223,6 +227,18 @@ static void cli_incoming_list_totals(str *instr, struct streambuf *replybuffer)
...
+		(unsigned long long)offers_ps.ps_min,
+		(unsigned long long)offers_ps.ps_max,
+		(unsigned long long)offers_ps.ps_avg);
+	streambuf_printf(replybuffer, " Min/Max/Avg answer requests per second          :%llu/%llu/%llu per sec\n",	(unsigned long long)answers_ps.ps_min,
+		(unsigned long long)answers_ps.ps_max,
+		(unsigned long long)answers_ps.ps_avg);
+	streambuf_printf(replybuffer, " Min/Max/Avg delete requests per second          :%llu/%llu/%llu per sec\n",
+		(unsigned long long)deletes_ps.ps_min,
+		(unsigned long long)deletes_ps.ps_max,
+		(unsigned long long)deletes_ps.ps_avg);
```
rfuchs: Aren't these missing the division `avg / count` ? Or is this intentional?

daemon/cli.c
claudiupb: `rtpe_totalstats_lastinterval` is just o copy of `rtpe_totalstats_interval` that is made once per graphite interval. When calling `send_graphite_data`, the stats are sent to graphite but they are also copied to a new struct that is later copied to `rtpe_totalstats_lastinterval`:
https://github.com/sipwise/rtpengine/blob/00bdcef9a4cd677d0c1ac36d83e4eed35917ef0c/daemon/graphite.c#L345-L353
This copy should be protected by the rtpe_totalstats_lastinterval_lock but it is not. This is an issue which I will address. 


The mutexes for each struct are used when the copy from `rtpe_totalstats_interval` is made so they don't need to be used here:
https://github.com/sipwise/rtpengine/blob/00bdcef9a4cd677d0c1ac36d83e4eed35917ef0c/daemon/graphite.c#L45-L55

daemon/cli.c
claudiupb: This is intentional, the average is computed and saved in the `send_graphite_data` function. Dividing again by the count will mean that it is done twice and it is incorrect. This is also happening for the statistics for the average time for offer/answer/delete. I will address this issue in a new commit

claudiupb:
Hello @rfuchs, sorry for the late reply I was in vacation until today. I saw your comments and I will make new commits to this branch to solve the existing issues.

claudiupb:
I have added 3 new commits to this branch 
3850e99 addreses the get_set issue
74cecc0 fixes the lock usage for statistics. 
As I explained in the https://github.com/sipwise/rtpengine/pull/446#discussion_r163821126, `rtpe_totalstats_lastinterval` is just a copy of `rtpe_totalstats_interval`, so the counters are not actually increased here. This is why `rtpe_totalstats_lastinterval_lock` is sufficient when working with data inside it. However I noticed that when the copy is made, another lock is used so I changed this. I also noticed that the mutexes for the `request_time` are never initialized but they are used, so this was also added.

0ffb9e6 fixes the display of average request time when issuing rtpengine-ctl list totals, since the average is divided twice by the count

daemon/call.c
```diff
@@ -477,6 +495,7 @@ static void call_timer(void *ptr) {
 	struct rtp_stats *rs;
 	unsigned int pt;
 	endpoint_t ep;
+	int offers, answers, deletes;
```
rfuchs: These should probably be of type u_int64_t if you don't want to risk overflows.

claudiupb: I modified this, as this can become an issue if there are many calls

PR #455, opened 2018-02-07
Add primary-interface-group parameter

claudiupb:
This commit adds a new parameter to the config `--primary-interface-group`. With this a user can define a group of interfaces that will be used for calls where no `direction` is given, instead of using just the first interface. 

This is useful to extend the maximum number of calls (with no `direction`)  an rtpengine instance can handle since there will be more available ports.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,7 @@ option and which are reproduced below:
 	  -t, --table=INT                  Kernel table to use
 	  -F, --no-fallback                Only start when kernel module is available
 	  -i, --interface=[NAME/]IP[!IP]   Local interface for RTP
+	  --primary-interface-group=NAME; NAME Interface group used as default
 	  -l, --listen-tcp=[IP:]PORT       TCP port to listen on
 	  -u, --listen-udp=[IP46:]PORT     UDP port to listen on
 	  -n, --listen-ng=[IP46:]PORT      UDP port to listen on, NG protocol
@@ -252,6 +253,15 @@ The options are described in more detail below.
 
 	Specifies a local network interface for RTP. At least one must be given, but multiple can be specified.
 	See the section *Interfaces configuration* just below for details.
+
+* --primary-interface-group
+
+	Specifies a list of interfaces that would be used if the `direction` option isn't given.
+	The interfaces are selected in round-robin fashion in the order they are declared by the
+	`--interfaces` option.
+
+	If this is parameter is not given then the legacy behavior applies and the first interace will
+	be used. See the section *Interface configuration* for more details.
 
 * -l, --listen-tcp, -u, --listen-udp, -n, --listen-ng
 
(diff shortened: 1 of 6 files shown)


rfuchs:
Hi,

Just a quick question here, as I'm not sure whether you're aware that there's already a round-robin interface selection algorithm in place, and it looks like there's some duplicated logic here.

From your README:

```
For this option only the name of the interfaces must be given. They must correspond with the names defined
in the `--interface` option otherwise they will be ignored:

	interface = pub1/IPV4; pub1/IPV6; pub2/IPV4; pub2/IPV6; pub3/IPV4; pub3/IPV6;
	primary-interface-group = pub1; pub3

The interfaces from this group will be selected in round-robin fashion in according to the desired address
family for the call.
```

I think the same thing can be achieved by doing something like this:

```
	interface = pub:1/IPV4; pub:1/IPV6; pub2/IPV4; pub2/IPV6; pub:3/IPV4; pub:3/IPV6;
	primary-interface = pub
```

Note the additional colon in the interface name, i.e. `pub:1` instead of `pub1`. This creates a logical round-robin interface named `pub`, which if selected using `direction=pub` would do a round-robin selection between `pub:1` and `pub:3`. (Individual logical interfaces can still be selected with `direction=pub:1` etc).

There's some more info about this in the current README. Would this be sufficient for your use case, or do you require the additional logic from this patch?

pkuzak:
We weren't aware of the feature you've written. You mentioned a _primary-interface_ configuration option? What exactly does this do? 
In our use case the SIP Proxy does not need to have any knowledge about what logical interface names there are and which one to use depending on which media relay is going to be used. We currently have 10 media relays, each one having a different _interface_ configuration (the order of interfaces is different). The logical naming of interfaces is important for our redundancy concept.
Also, changing logical interfaces names should be transparent to our SIP Proxies.

rfuchs:
Sorry, I guess I need to clarify. The `primary-interface` option doesn't exist yet, but the round-robin selection algorithm does. From what I understand, doing the following should achieve the same as your patch would achieve:

* Rename your logical interfaces to the `name:suffix` schema, at least those you wish to use as a default/primary interface group.
* Implement a single `primary-interface` option, which takes the place of the `direction=` argument if no `direction` was specified.

Perhaps the `primary-interface` option can even be skipped or made optional by falling back to an interface called `default` (or something) if no direction is specified. If no `default` interface exists, then it falls back to the current behaviour (pick the first). Then your interfaces could simply be called `default:1`, `default:2`, etc and you would get your round-robin selection.

pkuzak:
Thanks for your explanation. I think your proposal should also fit our needs. In our active-active redundancy setup each of or media relays has a differt default or primary interface logical name. The reason for this is how sessions get stored in Redis. The logical interface name is used to determine on which interface to restore (or create upon Redis notification) the session. So we kind of need a unique mapping between logical name to an IP on all media relays.
In order to get rid of the _primary-interface_ option, would following be OK for you:
If no _direction_ parameter is specified and the 1st configured logical interface name follows your proposal (_name:id_), _rtpengine_ will round-robin using all interfaces with the same name.

rfuchs:
Sounds like a plan :+1:

claudiupb:
I am closing this Pull Request and will open a new one with the proposed solution.

PR #456, opened 2018-02-13
Extend rtpengine ctl

claudiupb:
This branch adds the following to `rtpengine-ctl`:
* add `get` command as alias for `list`, in order to be more intuitive since the parameters are changed with `set` command.
* make sure that every parameter that can be `set` can also be `get`, so split the `list timeout` command
* make `control-tos`, `redis-allowed-errors`, `redis-disable-time`, `redis-cmd-timeout` and `redis-connect-timeout` configurable via `rtpengine-ctl`


claudiupb:
Thank you, I didn't know this function exists. I will use it in the future.

PR #467, opened 2018-02-17
adding no-rtcp-filtering flag

jerzyptak:
Some SIP endpints rely on rtcp feedback messages despite usage of RTP/(S)AVP profile.
This change adds no-rtcp-filtering flag that disables rtcp feedback filtering on RTP/(S)AVP streams.

### DIFF
--- a/daemon/rtcp.c
+++ b/daemon/rtcp.c
@@ -799,7 +799,7 @@ int rtcp_payload(struct rtcp_packet **out, str *p, const str *s, int no_rtcp_fil
 	if (rtcp->header.version != 2)
 		goto error;
 	err = "invalid packet type";
-	if (rtcp->header.pt != RTCP_PT_SR && rtcp->header.pt != RTCP_PT_RR
+	if (rtcp->header.pt != RTCP_PT_SR && rtcp->header.pt != RTCP_PT_RR && rtcp->header.pt != RTCP_PT_SDES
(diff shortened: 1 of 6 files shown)


daemon/rtcp.c
```diff
@@ -799,7 +799,7 @@ int rtcp_payload(struct rtcp_packet **out, str *p, const str *s, int no_rtcp_fil
 	if (rtcp->header.version != 2)
 		goto error;
 	err = "invalid packet type";
-	if (rtcp->header.pt != RTCP_PT_SR && rtcp->header.pt != RTCP_PT_RR
+	if (rtcp->header.pt != RTCP_PT_SR && rtcp->header.pt != RTCP_PT_RR && rtcp->header.pt != RTCP_PT_SDES
```
rfuchs: What is this based on? AFAIK even with AVPF and RTCP feedback messages, the first RTCP type in a compound packet must always be either RR or SR.

rfuchs:
See commit above (4991c3c) for a possible implementation of this.

daemon/rtcp.c
jerzyptak: When I added it I was not sure if the compound packets are handled as one or as separate ones, so I have added it as RTCP_PT_SDES is still a valid rtcp payload.

rfuchs: This line only applies to the first RTCP component in a compound packet, so I believe it should be left unchanged. Unless I'm missing something here.

jerzyptak: I agree, if the change has no effect then there is no point for it.

jerzyptak:
Great I will try it out as soon as I can.

PR #472, opened 2018-02-23
Sdp parsing fix for fmtp attribute

jerzyptak:
I have noticed that parsing of parameters in fmtp lines was incorrectly handling spaces, it would stop processing them at the first space. This does not look like a valid behaviour as fmtp formatting is dependent on the codec. For example this H.263 fmtp parameters set is valid
`a=fmtp:34 VGA=2;CIF=1;QVGA=1;QCIF=1;XMAX=640 YMAX=480 MPI=2;MaxBR=7680` 
but rtpengine will copy it as:
`a=fmtp:34 VGA=2;CIF=1;QVGA=1;QCIF=1;XMAX=640`

I have implemented fix for this issue.

### DIFF
--- a/daemon/sdp.c
+++ b/daemon/sdp.c
@@ -273,6 +273,18 @@ static int parse_address(struct network_address *address) {
 			&address->address_type, &address->address);
 }
 
+INLINE int extract_rest(char **sp, char *end, str *out) {
+	out->s = *sp;
+
+	if (end == *sp)
+		return -1;
+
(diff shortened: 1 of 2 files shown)


rfuchs:
Ya, good catch. The whole token extraction thing should actually be done with str_token() and str_token_sep(), eliminating the extract_token() function altogether. I'll fix it.

rfuchs:
19345d9 should be a clean fix for this

PR #517, opened 2018-04-07
 recording-daemon: added possibility to store audio streams to database

juha-h:
- added new column 'stream mediumblob' to 'recording_streams' table for audio streams
- added new option 'output-storage' with values 'file' (= default), 'db' or 'both'


rfuchs:
I've amended it slightly so that people who are already using the existing DB schema don't have to update it if they're not using the `db` output option. See commit af963a23f

juha-h:
Richard Fuchs writes:

> I've amended it slightly so that people who are already using the
> existing DB schema don't have to update it if they're not using the
> `db` output option. See commit af963a23f

I too had that in mind, i.e., use stream field only if storage mode is
'db' or 'both'.

-- Juha

PR #534, opened 2018-04-25
ffmpeg: handle some API changes for release/4.0 master

aalba6675:
Handle some API changes for newer versions of ffmpeg.

Encountered with Fedora 28 + latest ffmpeg

### DIFF
--- a/lib/fix_frame_channel_layout-01.h
+++ b/lib/fix_frame_channel_layout-01.h
@@ -5,5 +5,9 @@
 INLINE void fix_frame_channel_layout(AVFrame *frame) {
 	if (frame->channel_layout)
 		return;
+#if LIBAVUTIL_VERSION_MAJOR >= 56
+	frame->channel_layout = av_get_default_channel_layout(frame->channels);
+#else
 	frame->channel_layout = av_get_default_channel_layout(av_frame_get_channels(frame));
+#endif
 }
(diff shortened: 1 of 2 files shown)


lib/fix_frame_channel_layout-01.h
```diff
@@ -5,5 +5,9 @@
 INLINE void fix_frame_channel_layout(AVFrame *frame) {
 	if (frame->channel_layout)
 		return;
+#if LIBAVUTIL_VERSION_MAJOR >= 56
+	frame->channel_layout = av_get_default_channel_layout(frame->channels);
+#else
 	frame->channel_layout = av_get_default_channel_layout(av_frame_get_channels(frame));
+#endif
```
rfuchs: The purpose of the different `fix_frame_channel_layout-*.h` is actually to eliminate the dependency on version test macros. (There seem to be too many different libav/ffmpeg versions out there, with inconsistent version numbers, that I gave up on finding the correct version test for this particular code.) If the `-01.h` fails to compile, it's supposed to move on to `-02.h` and use that if it compiles. Are you sure this is actually needed?

aalba6675: I missed that:  - -01.h actually compiles only the compiler was issuing deprecation warnings.

This version check can be omitted as the code still works.

rfuchs:
See ae3dae7 for an alternative, hopefully handling all cases. Let me know how that works for you.

rfuchs:
Let's see how long it takes before somebody complains that this broke something :)

PR #548, opened 2018-05-10
allow empty local address in interface config if advertised address is specified

ghost:
In our case, we want to receive rtp packets from all interfaces which all local interfaces have the same advertised address. Instead of listing all interfaces in config, an implicit 'all local addresses' way is preferred.
The real receiving interface is learned from the first received packet to make kernel forwarding works properly.

allowed format:
interface = !192.168.1.1

### DIFF
--- a/daemon/socket.c
+++ b/daemon/socket.c
@@ -3,6 +3,7 @@
 #include <unistd.h>
 #include <errno.h>
 #include <netdb.h>
+#include <netinet/in.h>
 #include <netinet/ip.h>
 #include <netinet/ip6.h>
 #include <netinet/udp.h>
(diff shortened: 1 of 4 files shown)


daemon/socket.c
```diff
@@ -320,7 +325,28 @@ static ssize_t __ip_recvfrom_ts(socket_t *s, void *buf, size_t len, endpoint_t *
...
 				*tv = *((struct timeval *) CMSG_DATA(cm));
 				tv = NULL;
-				break;
+			} else if (cm->cmsg_level == IPPROTO_IP && cm->cmsg_type == IP_PKTINFO) {
+				struct in_pktinfo *pktinfo = (struct in_pktinfo *)CMSG_DATA(cm);
+
+				if (is_addr_unspecified(&s->local.address))
+					ilog(LOG_WARNING, "Socket local address is already specified");
+
+				s->local.address.u.ipv4 = pktinfo->ipi_spec_dst;
```
rfuchs: There's a potential race condition here as multiple threads might be accessing the s->local struct simultaneously.  Originally this was supposed to be filled during socket creating and then be read-only afterwards, making thread synchronization unnecessary.

PR #578, opened 2018-07-17
Update README.md

kosza:
In Ubuntu Bionic the xtables are moved to different path.
The new kernel modul available after system reboot.


rfuchs:
The changed path actually applies to newer Debian releases as well, and I assume also some other distros.

I'm not sure about that reboot statement. Why would a reboot be needed?

kosza:
It is a good question. Honestly I don't know.
modprobe didn't see xt_RTPENGINE.ko. For test we rebooted the system and voila.
It seems that the Bionic needs reboot. Later we will do a clear install and we can test it again.

PR #586, opened 2018-08-02
Make recording more accurately report source address

egreenmachine:
Just as a note. I have tested the pcap recording method, but not the proc recording method. I tried to make the commit as unintrusive as possible and I expect that proc recording works as it should.

### DIFF
--- a/daemon/recording.c
+++ b/daemon/recording.c
@@ -40,7 +40,7 @@ static char *meta_setup_file(struct recording *recording);
 static int pcap_create_spool_dir(const char *dirpath);
 static void pcap_init(struct call *);
 static void sdp_after_pcap(struct recording *, GString *str, struct call_monologue *, enum call_opmode opmode);
-static void dump_packet_pcap(struct recording *recording, struct packet_stream *sink, const str *s);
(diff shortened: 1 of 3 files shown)


daemon/recording.c
```diff
@@ -490,9 +490,9 @@ static void pcap_recording_finish_file(struct recording *recording) {
 }
 
 // "out" must be at least inp->len + MAX_PACKET_HEADER_LEN bytes
-static unsigned int fake_ip_header(unsigned char *out, struct packet_stream *stream, const str *inp) {
-	endpoint_t *src_endpoint = &stream->advertised_endpoint;
-	endpoint_t *dst_endpoint = &stream->selected_sfd->socket.local;
+static unsigned int fake_ip_header(unsigned char *out, struct media_packet *mp, const str *inp) {
+	endpoint_t *src_endpoint = &mp->fsin;
+	endpoint_t *dst_endpoint = &mp->stream->selected_sfd->socket.local;
```
rfuchs: You should be able to use `mp->sfd->socket.local` here

egreenmachine:
Fixed

PR #658, opened 2018-11-16
Bria softphone using SRTP call with AES_CM_256_HMAC_SHA1_80 requires padding to append

adil-mafzool:
crypto containing padding should be sent within sdp for pass through

### DIFF
--- a/daemon/call_interfaces.c
+++ b/daemon/call_interfaces.c
@@ -36,7 +36,7 @@ static pcre_extra *streams_ree;
 
 int trust_address_def;
 int dtls_passive_def;
-
+int pad_crypto_def;
 
 static int call_stream_address_gstring(GString *o, struct packet_stream *ps, enum stream_address_format format) {
 	int len, ret;
@@ -621,6 +621,8 @@ static void call_ng_flags_flags(struct sdp_ng_flags *out, str *s, void *dummy) {
 		out->always_transcode = 1;
 	else if (!str_cmp(s, "asymmetric-codecs"))
 		out->asymmetric_codecs = 1;
+	else if (!str_cmp(s, "pad-crypto"))
+		out->pad_crypto = 1;
 	else {
 		// handle values aliases from other dictionaries
 		if (call_ng_flags_prefix(out, s, "SDES-", ng_sdes_option, NULL))
@@ -654,6 +656,7 @@ static void call_ng_process_flags(struct sdp_ng_flags *out, bencode_item_t *inpu
 
 	out->trust_address = trust_address_def;
 	out->dtls_passive = dtls_passive_def;
+	out->pad_crypto = pad_crypto_def;
 
 	call_ng_flags_list(out, input, "flags", call_ng_flags_flags, NULL);
 	call_ng_flags_list(out, input, "replace", call_ng_flags_replace, NULL);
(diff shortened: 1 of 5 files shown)


adil-mafzool:
Bria softphone using SRTP call with AES_CM_256_HMAC_SHA1_80 requires padding when using the base64 string you are using there is not padded; its length is not a multiple of 3. From RFC 4648:
Implementations MUST include appropriate pad characters at the end of encoded data unless the specification referring to this document explicitly states otherwise.

The call make from Bria soft phone fails with following bria error.
"SDES crypto key found in SDP, but is not of correct length after base 64 decode: 45, should be: 46"

AES_CM_256_HMAC_SHA1_80
Bria : inline:IPX7TGb7HLAYBGuuGdP//+VVSLFV4cz8WrkWVvbA39BJGHk580/UrV+ioCRFiA==
RTPEngine: inline:CYkq/gl8W1Bik8xvqlmJxnznqfDD5fGg13aLsjTOku/ILo8K3RgOQaxlq1KYFe
RTPEngine version: 6.5.2.0+0

rfuchs:
Yes, and the document that does refer to RFC 4648 in this particular case is RFC 4568, which states in section 6.1:
```
When base64 decoding the key and
   salt, padding characters (i.e., one or two "=" at the end of the
   base64-encoded data) are discarded (see [RFC3548] for details).
```
I'm afraid the Bria phone is to blame here.

adil-mafzool:
@rfuchs The modified code worked fine. Please can you review the code and add to mr6.5 branch.

rfuchs:
No because while it might work for you, it might break things in other cases. My understanding of the RFC is that there should be no padding characters.

rfuchs:
That's your assumption. It's just as well possible that a receiver will complain about extra padding characters being present when they shouldn't be, just as Bria complains about them not being present when it's expecting them to be there.

Unless you have any information to the contrary, I maintain that the RFC states that the padding characters should be discarded, and that Bria is wrong in rejecting the attribute.

adil-mafzool:
@rfuchs digging Rfc4648 states section 3.2 : -

`Implementations MUST include appropriate pad characters at the end of encoded data unless the specification referring to this document explicitly states otherwise.

The base64 and base32 alphabets use padding, as described below in sections 4 and 6, but the base16 alphabet does not need it; see section 8.`

Base64 section 4 relating with padding : -

`Padding at the end of the data is performed using the '=' character.  Since all base 64 input is an integral number of octets, only the following cases can arise:

(1) The final quantum of encoding input is an integral multiple of 24 bits; here, the final unit of encoded output will be an integral  multiple of 4 characters with no "=" padding.

(2) The final quantum of encoding input is exactly 8 bits; here, the final unit of encoded output will be two characters followed by two "=" padding characters.

 (3) The final quantum of encoding input is exactly 16 bits; here, the final unit of encoded output will be three characters followed by one "=" padding character.`

rfuchs:
"unless the specification referring to this document explicitly states otherwise"

RFC 4568 does explicitly state otherwise.

rfuchs:
That being said, from looking at sample SDPs that can be found online, it seems that both variants (with and without padding) are commonly found. So I wouldn't mind having a compile-time option or something like that. But the default should be the behaviour as it is now.

adil-mafzool:
@rfuchs can you review the new flag `pad-crypto`

daemon/call_interfaces.c
```diff
@@ -36,7 +36,7 @@ static pcre_extra *streams_ree;
 
 int trust_address_def;
 int dtls_passive_def;
-
+int pad_crypto_def;
```
rfuchs: I don't see a way to set this to anything other than the default 0? I think this can be omitted unless you also want to add a global config option (in which case it should go into rtpe_config).

adil-mafzool: removed the unnecessary pad_crypto_def variable

adil-mafzool: Will it be possible to merge into mr6.5 branch?

rfuchs:
Merged to mr6.5

Looks like Bria has the same problem with FreeSwitch: https://freeswitch.org/jira/si/jira.issueviews:issue-html/FS-10864/FS-10864.html

PR #690, opened 2019-01-25
Work around ssrc context for handling srtp replay protection for the transcoded srtp stream

adil-mafzool:
Work around ssrc context for handling srtp replay protection for the transcoded srtp stream


rfuchs:
I'm working on a similar fix, but more elaborate and extensive to provide consistent seq and TS output in all cases.

PR #731, opened 2019-03-11
Fix lower or higher target CentOS Kernel Versions

khorsmann:
The changes in https://github.com/sipwise/rtpengine/commit/62ec9cc4d28fafac6a4937dccebfe2f852a77aec breaks the installation of ngcp-rtpengine-dkms. It will not produce any kind of xt_RTPENGINE.* module if your kernel-version dont match with the build-host. Only chance is if you have the *same* kernel-version as the build-host.
More is less here. Remove this changes.

### DIFF
--- a/el/rtpengine.spec
+++ b/el/rtpengine.spec
@@ -72,10 +72,6 @@ BuildRequires:  gcc make redhat-rpm-config mysql-devel ffmpeg-devel
 %define binname rtpengine
 %define archname rtpengine-mr
 
-%{!?kversion: %define kversion %(uname -r)}
-# hint: this can be overridden with "--define kversion foo" on rpmbuild,
-# e.g. --define "kversion 2.6.32-696.23.1.el6.x86_64"
-
 %prep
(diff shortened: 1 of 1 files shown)


rfuchs:
You'll have to discuss with @netaskd 

Without any more details, I would strongly favour strict version-specific kernel module builds. Unless the RPM platform does something different here, the kernel itself makes zero guarantees about ABI compatibility even among patch releases.

netaskd:
@khorsmann please test c73c874

khorsmann:
@netaskd
#734 didnt work.
Better we use %if 0%{?kversion:1} like @rfuchs described with Professor Google.
I tested %if 0%{?kversion:1} and omited the previous declaration %if statement. Then it works fine. with and without define.

khorsmann:
@netaskd  last commit on #734 fix that 4cdf106.
Plus optional --define for (uname -r) on buildhost or (best kernelversion ever). I close my pull-request then.

PR #734, opened 2019-03-14
rtpengine.spec: fix building dkms module on a different host kernel ver

netaskd:
fixes commit 62ec9cc
In case when DKMS package built without defined kversion, it must build module for target host current kernel version.

### DIFF
--- a/el/rtpengine.spec
+++ b/el/rtpengine.spec
@@ -72,9 +72,6 @@ BuildRequires:  gcc make redhat-rpm-config mysql-devel ffmpeg-devel
 %define binname rtpengine
 %define archname rtpengine-mr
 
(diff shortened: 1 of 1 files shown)


netaskd:
@khorsmann please test the latest force push 4cdf106 It works perfect for the both cases.

rfuchs:
Thanks for taking care of this @netaskd

kdoren:
@netaskd I tried referencing kversion and install failed on target with different kernel.  I wound up creating my own rpm specfile based on fedora specfile from @amessina  (see [(https://github.com/sipwise/rtpengine/issues/758)] ), eliminated the kversion references, and it works fine installing on target with different kernel.   Differing/changing kernel  versions is the problem that DKMS was designed to solve, it works automatically.

PR #738, opened 2019-03-17
package files reorganization

### LINKED ISSUE #737: organize packaging files in order to easily support more versions and flavours
It would be nice to have all packaging files organized in a way that automated tools can be used to build them.

My proposal is, as We do in kamailio project: having a ``pkg/`` directory with all the supported versions. ``pkg/deb/debian`` would be the sid version and some scripts for each supported version/flavor at ``pkg/deb/debian/backports`` that will produce ``pkg/deb/$(version)``

So the build process would need a previous step: ``ln -s pkg/deb/<codename> debian``

linuxmaniac:
define stretch as first supported deb version

Fixes #737


taurus-forever:
It is looks line necessary changes if we want to build rtpengine for several .deb-based distributions.
I am not sure how to keep in sync files into pkg/deb and debian/* after changes in branches.
Sure for the latest one you will have symlinks... but still some changes can happens in NGCP branch mr5.5 which should be represented in pkg/deb/jessie/control, etc.

LGTM as it doesn't affect NGCP releasing.

Up 2 @mika and @guillemj  here.

P.S. @linuxmaniac TNX!

mika:
Hej @linuxmaniac - thanks for taking care! :) So we briefly discussed this internally, we're not _yet_ sure whether that's the best approach for us, as it seems to go in the wrong direction, instead of removing all packaging it adds even more. :)

@linuxmaniac, do we have an overview of the current state of affairs of all the Debian + Ubuntu releases we (as in: building and providing packages at kamailio.sipwise.com) are targeting? (We could always special case rtpengine and guarantee it is buildable with oldstable and stable, possibly also testing.)

Another option might be to use master with what we have, and use debian/$release and ubuntu/$release or so (following https://dep-team.pages.debian.net/deps/dep14/) for their specific packaging. (We could try to automatically pull and rebase the branches against master during packaging builds?) Does anyone have experience with a setup like that?

linuxmaniac:
@mika I didn't even started to try to build on all the releases that kamailio 5.3 supports, so no for now I don't know the current stat of affairs

From my POV you don't even need to maintain pkg/deb/debian/backports/* or any pkg/whatever. It should be a community effort.

rfuchs:
As of right now, I can't merge this because our jenkins-debian-glue refuses to build it due to lack of /debian/

taurus-forever:
Hi @linuxmaniac,

As discussed on the "Courage meeting" we need to improve this a bit.
1) it is fine to have "pkg" folder with all the supported suites.
2) you need to move folder "el" into "pkg/rhel/el5" to be solid. Right? Maybe fire a separate pull request and ask RedHat files contributors to check it.
3) Move current "debian/" to "pkg/debian/stretch/" and create "debian/" symlink to "pkg/debian/stretch/" , so those changes will be fully compatible with Sipwise. Later on during migration to buster, we will create "pkg/debian/buster" and update symlink.

taurus-forever:
@linuxmaniac  resolved/abandoned with https://github.com/sipwise/rtpengine/commit/bbfe168b7201c7bec49fca7b301956daa365bd2a ?

PR #740, opened 2019-03-21
add kernel-devel package as requires for DKMS module build on a target host

netaskd:
DKMS module will not builded without kernel sources on the target host


mika:
AFAIK the `kernel-devel` package is rhel + fedora specific, but fails e.g. on SuSE, maybe we should use an approach like ZOL folks in https://github.com/zfsonlinux/zfs/blob/ecd3728b264b94de239932555b7046f1ddb936b2/rpm/generic/zfs-kmod.spec.in#L75?

netaskd:
Hi @mika,
Thats a good idea! Changed in c2ad397

mika:
Thanks for taking care, @netaskd - LGTM!

PR #755, opened 2019-04-16
Insert streambuf object to listener->streams before adding it to poller

vitalikvoip:
If a CLI client disconnects immediately after connection establishment rtpengine sometimes doesn't remove resources (struct streambuf_stream) due to the race between poller threads (the one that accepted connection and another one that closed it).

To reproduce:
run: ./cli_test 1000
[cli_test.zip](https://github.com/sipwise/rtpengine/files/3085077/cli_test.zip)

Some sockets will stay in CLOSE_WAIT forever.


rfuchs:
I think this might add another race condition via the refcount of `s`. It would need an extra `obj_hold` on it before being added to the `streams` hash table, followed by an `obj_put` at the end of the function.

vitalikvoip:
Checked the code and didn't understand what can go wrong between adding `obj` to `streams` and insertions of a new `poller_item` to a poller. Could you pls describe the case.

rfuchs:
A concurrent `streambuf_stream_closed` event releases a reference of `s` if it was present in the `streams` hash table. If this happens right before `poller_add_item` is called, then `s` would be freed while `i` still points to it, before `poller_add_item` can grab a reference to it.

I'm not sure if this can actually happen with the current code base, but for robustness's sake the possibility should be accounted for.

vitalikvoip:
Well `streambuf_stream_closed` is called from event callbacks which are invoked by a poller, so nobody can call them before `streambuf_stream` is added to poller.
But I see your point an additional guarding `obj_hold`/`obj_put` won't harm.
I will add them and update the pull request.

vitalikvoip:
Added additional `obj_hold`/`obj_put` around `streambuf_stream` object.

daemon/tcp_listener.c
```diff
@@ -172,20 +172,25 @@ static void streambuf_listener_newconn(struct obj *p, socket_t *newsock, char *a
...
+	if (poller_add_item(listener->poller, &i))
+		goto fail;
+
+	obj_put(s);
+
 	return;
 
 fail:
 	obj_put(s);
+	obj_put(s);
```
rfuchs: This one should not be necessary. The `fail` label is only reached after `s` has been added to `streams` which hands over one reference. So you're left with only one reference to release.

These changes should be squashed into a single commit. I can do that and submit it if you prefer.

vitalikvoip:
Squashed commits. 
If I understood the code correctly this additional `obj_put` in the end is needed. Objects added to `streams` are cleaned from `streambuf_stream_closed` which in its turn is called from different callbacks like `streambuf_stream_readable` or `streambuf_stream_writeable`. If we failed to add an item to poller those callbacks will never be called and thus `streams` will stay in `streams` forever.

After writing all this have a feeling that this patch is not very good. If we fail to add to poller my patch will remove `streambuf_stream` object but it will stay as a dangling pointer in `listener->streams`.

rfuchs:
Hm yes you're right. I guess the code needs to clean up `streams` in the failure path.

vitalikvoip:
Updated pull request.

PR #831, opened 2019-08-26
EL8: mariadb-devel conflicts with mysql-devel

space88man:
Detect EL8 and use mariadb-devel instead.


space88man:
In EL8 , `mariadb-devel` does not provider `mysql-devel`, but instead conflicts with it. :cry:

Issue #803, opened 2019-07-10
Jitter Buffer Design For RTPEngine


balajee-plivo:
Hello All,

We had [opened a feature request](https://github.com/sipwise/rtpengine/issues/785) for implementing Jitter Buffer for RTPEngine. 

We have been working on this and we have come up with a design for the Jitter Buffer. The development of the Jitter Buffer is underway. We thought we would publish the design and open it up for feedback and comments from the community. 

The design is captured in [this gist here](https://gist.github.com/balajeesv/1a0cec2a92ed6eb5299e56bd1691457b) 

Also, apologies for creating an issue for this as we weren't sure where to post such a request.  If this isn't the right forum to capture feedback, please do let me know how to proceed with this. 

Thanks!
Balajee

rfuchs:
Hi,

Thanks for your interest and work so far. My initial thoughts regarding your proposed design:

For media playback purposes, rtpengine already has a timer thread pool, which works more or less the same as your proposed jitter buffer timer threads. I see no reason why this cannot be used for a generic jitter buffer (and in fact this is what I would have used). Look for `send_timer` in the sources, in particular in `media_player.c` and the low level implementation in `timerthread.c`. It doesn't use a timerfd with a fixed tick length, but otherwise handles the scheduling of RTP packets to send out just fine.

The `send_timer` currently operates by converting the RTP timestamp into a `timeval` timestamp, which is then used to determine when to send the packet out. This is different from using the ptime, but (IMO) is more resilient to ptime mismatches as it requires no knowledge of the media format used in the RTP payloads. All that's needed is the clock rate used for the stream, which is known from the SDP. The disadvantage is that RTP timestamp resets must be handled. (Other considerations not taken into account yet are: system clock drifts, which can be fixed by using a different time source; and possibly RTP sequence numbering in case there's multiple packets with the same RTP timestamp.)

So most of the work that's needed for a jitter buffer is already done. What's needed is that for normal stream relay (not transcoding, not playing back media) when the option is enabled, instead of just sending the RTP packets back out again, to pass them to the same RTP processing functions that handle media playback and thus automatically benefit from the jitter-free playback routine.

That being said, there's two crucial particulars missing.

[...]

PR #834, opened 2019-08-27
Jitter Buffer implementation for rtpengine

balajeesv:
Hi,
  Please review Jitter buffer implementation in rtpengine. Discussions regarding jitter buffer implementation can be found [here](https://github.com/sipwise/rtpengine/issues/803).

The algorithm is mentioned in the [gist](https://gist.github.com/balajeesv/1a0cec2a92ed6eb5299e56bd1691457b) and the timer is implemented with existing timer thread pool instead of ptime proposed. 

The existing timer is enhanced to support jitter buffer by introducing a few changes like insert into the queue with sorting in descending order (sorted in descending order for faster insertion) and read from the tail. 

Tested this code with basic buffering, packets drops, duplicates, repacketization, and transcoding.

Thanks,
Balajee

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -290,8 +290,11 @@ retry:
 		sigprocmask(SIG_SETMASK, &ss, NULL);
 		closelog();
 
-		for (i = 0; i < 100; i++)
+		for (i = 0; i < 100; i++) {
+			if (i == 2 && rtpe_config.common.log_stderr)
+				continue;
 			close(i);
+		}
 
 		if (!rtpe_config.common.log_stderr) {
 			openlog("rtpengine/child", LOG_PID | LOG_NDELAY, LOG_DAEMON);
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -899,11 +899,13 @@ struct packet_stream *__packet_stream_new(struct call *call) {
 	stream = uid_slice_alloc0(stream, &call->streams);
 	mutex_init(&stream->in_lock);
 	mutex_init(&stream->out_lock);
+	mutex_init(&stream->jb.lock);
 	stream->call = call;
 	atomic64_set_na(&stream->last_packet, rtpe_now.tv_sec);
 	stream->rtp_stats = g_hash_table_new_full(g_int_hash, g_int_equal, NULL, __rtp_stats_free);
 	recording_init_stream(stream);
-	stream->send_timer = send_timer_new(stream);
+	stream->send_timer = send_timer_new(stream, 0);
+	stream->buffer_timer = send_timer_new(stream, 1);
 
 	return stream;
 }
@@ -2166,6 +2168,8 @@ void call_destroy(struct call *c) {
 				ps = o->data;
 
 				send_timer_put(&ps->send_timer);
+				send_timer_put(&ps->buffer_timer);
+				mutex_destroy(&ps->jb.lock);
 
 				if (PS_ISSET(ps, FALLBACK_RTCP))
 					continue;
@@ -2367,6 +2371,7 @@ static struct call *call_create(const str *callid) {
 	c->dtls_cert = dtls_cert();
 	c->tos = rtpe_config.default_tos;
 	c->ssrc_hash = create_ssrc_hash_call();
+	c->enable_jb = rtpe_config.enable_jb;
 
 	return c;
 }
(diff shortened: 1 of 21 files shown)


balajeesv:
With new commit sender clock drift handling and few other minor fixes are done.Please check.

balajee-plivo:
@rfuchs  Did you get chance to review this PR

balajeesv:
@rfuchs Did you get a chance to review this PR?

rfuchs:
No not yet, sorry for the delay, I'll try to get to it one of these days

daemon/call.c
```diff
@@ -899,11 +899,13 @@ struct packet_stream *__packet_stream_new(struct call *call) {
...
 	mutex_init(&stream->in_lock);
 	mutex_init(&stream->out_lock);
+	mutex_init(&stream->jb.lock);
 	stream->call = call;
 	atomic64_set_na(&stream->last_packet, rtpe_now.tv_sec);
 	stream->rtp_stats = g_hash_table_new_full(g_int_hash, g_int_equal, NULL, __rtp_stats_free);
 	recording_init_stream(stream);
-	stream->send_timer = send_timer_new(stream);
+	stream->send_timer = send_timer_new(stream, 0);
+	stream->buffer_timer = send_timer_new(stream, 1);
```
rfuchs: These should be either different functions (send_timer_new and jitter_buffer_new or something like that), or only use a single timer object.

daemon/media_socket.c
```diff
@@ -1677,7 +1678,7 @@ static int do_rtcp(struct packet_handler_ctx *phc) {
 int media_socket_dequeue(struct media_packet *mp, struct packet_stream *sink) {
 	struct codec_packet *p;
 	while ((p = g_queue_pop_head(&mp->packets_out)))
-		send_timer_push(sink->send_timer, p);
+		send_timer_push((p->packet && p->packet->buffered)?sink->buffer_timer:sink->send_timer, p);
```
rfuchs: Should be able to eliminate this case distinction.

daemon/jitter_buffer.c
```diff
@@ -0,0 +1,335 @@
...
+#define CONT_INCORRECT_BUFFERING 0x14
+
+
+static struct jitter_buffer_config *jb_config; 
+
+struct jitter_buffer_config {
+	int    min_jb_len;
+	int    max_jb_len;
+	int    enable_jb;
+	int    src_clockrate;
```
rfuchs: I'm a bit confused about this parameter as I don't really see it being used, other than as an enable switch for the clock drift calculation.

daemon/jitter_buffer.c
```diff
@@ -0,0 +1,335 @@
...
+		(long long) (ts_diff)* 1000000 / clockrate;
+	struct timeval to_send = jb->first_send;
+	timeval_add_usec(&to_send, ts_diff_us);
+	long long time_diff = timeval_diff(&rtpe_now, &to_send);
+
+	jb->clock_drift_val = time_diff/seq_diff;
+	jb->clock_drift_enable = 0;
+	jb->cont_buff_err = 0;
+}
+
```
rfuchs: Can you explain the algorithm used here? AFAICT this is a single-shot calculation that is done when an overflow or underflow occurs?

balajeesv:
> Ok, so from a quick first glace, I see that this is implemented as an incoming packet buffer, unlike the send_timer mechanism that already exists, which is an outgoing packet buffer. Originally I was under the impression that this would be another outgoing packet buffer, which is why I suggested using the send_timer.
> 
> This makes me think that a more generic version of send_timer would be more appropriate in order to eliminate some unnecessary case distinctions. Either completely unlinked from the send_timer as another instance of a timerthread (which would require separate dedicated threads), or within the send_timer but using a handler function in codec_packet (which then should probably be renamed).
> 
> I can probably handle this kind of refactoring if you prefer.

Incoming packet buffer used because of transcoding and re-packetization would get benefitted from the incoming buffer.
I would be happy to get this refactored to a more generic version of send_timer :)

balajee-plivo:
@rfuchs Did you get chance to look into responses?

rfuchs:
Yes, thank you, I plan on getting on this shortly.

rfuchs:
Just a quick update here to let you know that I'm almost through with refactoring and integrating your code.

rfuchs:
I've pushed my refactored version of this into the branch `rfuchs/jb`. If you can, please try it and see if/how it works for you.

balajee-plivo:
Sure. Will test and update you.

balajeesv:
Sure. I will test and update you.

PR #836, opened 2019-08-29
spandsp-devel is missing

denyspozniak:
added `spandsp-devel` to the build requirements.


rfuchs:
Can you squash those please :laughing:

PR #856, opened 2019-10-01
Add missing errno.h

xadhoom:
Fixes the following compilation error:

dtmf.c: In function 'dtmf_init':
dtmf.c:15:77: error: 'errno' undeclared (first use in this function)
    ilog(LOG_ERR, "Failed to open/connect DTMF logging socket: %s", strerror(errno));

### DIFF
--- a/daemon/dtmf.c
+++ b/daemon/dtmf.c
@@ -7,6 +7,7 @@
 #include "rtplib.h"
 #include "codec.h"
 #include "ssrc.h"
+#include "errno.h"
 
 
 


daemon/dtmf.c
```diff
@@ -7,6 +7,7 @@
 #include "rtplib.h"
 #include "codec.h"
 #include "ssrc.h"
+#include "errno.h"
```
rfuchs: That should be <errno.h> instead of "errno.h"

PR #895, opened 2019-12-14
RT-985 Fix issue if period of silence is longer than 5 seconds

benlai2018:


rfuchs:
Way too large of a PR with all kinds of unrelated changes and no description provided.

Issue #919, opened 2020-01-30
RTPEngine Crashes on redis pubsub connection loses.


pateldp:
I am using rtpengine with redis pub sub support with haproxy for redis connection.

When i'm connecting redis server directly then there is no issue at all but when i put haproxy between them. I got following errors continuously and rtpengine crashes.

Jan 30 02:23:41 olp-gen-sbca02 rtpengine[6626]: ERR: redis_async_context_disconnect error 3 on context free: Server closed the connection
Jan 30 02:23:41 olp-gen-sbca02 rtpengine[6626]: ERR: Lost connection to Redis 10.100.2.102:6378
Jan 30 02:23:41 olp-gen-sbca02 rtpengine[6626]: INFO: Asking Redis whether it's master or slave...
Jan 30 02:23:41 olp-gen-sbca02 rtpengine[6626]: INFO: Connected to Redis 10.100.2.102:6378 in master mode
Jan 30 02:23:41 olp-gen-sbca02 rtpengine[6626]: INFO: RE-Established connection to Redis 10.100.2.102:6378
Jan 30 02:23:41 olp-gen-sbca02 rtpengine[6626]: INFO: Use Redis 10.100.2.102:6378 for notifications
Jan 30 02:23:41 olp-gen-sbca02 rtpengine[6626]: INFO: Use Redis 10.100.2.102:6378 to subscribe to notifications 

haproxy configuration
------------------------------------------------------------------
global
        maxconn 1024
        user haproxy
        group haproxy
        daemon
        stats socket /var/run/haproxy.sock mode 600 level admin         # Make sock file for haproxy
	log 0.0.0.0  local2 debug
	log /dev/log    local2 debug
 
defaults
        log     global
        mode    http
        option  tcplog
        option  dontlognull
        retries 3
        option  redispatch
        maxconn 10000
        timeout connect 5000ms
        timeout client 6s
        timeout server 6s
defaults REDIS
mode tcp
timeout connect 3s
timeout server 6s
timeout client 6s

# Specifies listening socket for accepting client connections using the default 
# REDIS TCP timeout and backend bk_redis TCP health check.
frontend ft_redis
bind *:6378 name redis
default_backend bk_redis

# Specifies the backend Redis proxy server TCP health settings 
# Ensure it only forward incoming connections to reach a master.
backend bk_redis
mode tcp
option tcp-check
tcp-check connect
tcp-check send PING\r\n
tcp-check expect string +PONG
[...]

Issue #10, opened 2014-07-21
parse_attribute_crypto


edjames99:
Hi All,
I think c->lifetime = 1 << c->lifetime should be c->lifetime = 1ULL << c->lifetime otherwise for a lifetime of say 2^31 the c->lifetime is populated with a negative value, presumably 1 is interpreted as signed by default.

Thanks
Ed James

PR #921, opened 2020-02-03
Fix crash when redis disconnects/reconnects

smititelu:
Steps to reproduce:
1. start rtpengine with redis support
2. stop the redis server
3. start the redis server => wait for rtpengine to reconnect => crash

From [1] I've found out that:
1. "The asynchronous context can hold a disconnect callback function that is called when the connection is disconnected (either because of an error or per user request)"
2. "The context object is always freed after the disconnect callback fired."

So I think there is no point in calling redisAsyncDisconnect() on redis reconnect; this will lead to the crash.

[1] https://github.com/redis/hiredis

### DIFF
--- a/daemon/redis.c
+++ b/daemon/redis.c
@@ -572,7 +572,6 @@ static int redis_notify(void) {
 }
 
 static void redis_disconnect(void) {
-	redisAsyncDisconnect(rtpe_redis_notify_async_context);
 	rtpe_redis_notify_async_context = NULL;
 }
 
@@ -634,6 +633,7 @@ void redis_notify_loop(void *d) {
 	redis_notify_subscribe_action(UNSUBSCRIBE_ALL, 0);
 
 	// free async context
+	redisAsyncDisconnect(rtpe_redis_notify_async_context);
 	redis_disconnect();
 }
 


rfuchs:
I assume this would fix #919 ?

smititelu:
I'm not using a haproxy configuration scenario, and my gdb trace looks a bit different:

(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007fd2bde4b535 in __GI_abort () at abort.c:79
#2  0x00007fd2bdea2508 in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7fd2bdfad28d "%s\n") at ../sysdeps/posix/libc_fatal.c:181
#3  0x00007fd2bdea8c1a in malloc_printerr (str=str@entry=0x7fd2bdfab43b "free(): invalid pointer") at malloc.c:5341
#4  0x00007fd2bdeaa42c in _int_free (av=<optimized out>, p=<optimized out>, have_lock=<optimized out>) at malloc.c:4165
#5  0x00007fd2be30a9f8 in __redisGetSubscribeCallback (dstcb=0x7fd2bb536820, reply=0x7fd29c005470, ac=0x7fd2bb537700) at async.c:479
#6  redisProcessCallbacks (ac=0x7fd2bb537700) at async.c:479
#7  0x000055a1af82160c in redis_disconnect () at redis.c:622
#8  redis_notify_loop (d=<optimized out>) at redis.c:622
#9  0x000055a1af813592 in thread_detach_func (d=0x55a1afddc3a0) at aux.c:207
#10 0x00007fd2bdff3fa3 in start_thread (arg=<optimized out>) at pthread_create.c:486
#11 0x00007fd2bdf224cf in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95


However, he can give it a try since I see it's about a lost connection and re-connection.

PR #988, opened 2020-05-10
Jb improvements

balajeesv:
1. JB correction on RTP Marker bit
2. Sequence wrap handling
3. Flush jitter buffer on reset
4. Ensure correct ordering of sequence
5. Comfort noise handling
6. Clock drift handling improvements


rfuchs:
For future pull requests, it would be appreciated if each commit had a more descriptive messages attached, detailing what it achieved, and also not "root" as author. Thank you.

PR #997, opened 2020-05-26
Add drop-traffic start and stop flags

smititelu:
Useful when someone needs to:
  - completely drop early media, but stop dropping after call is answered.
  - completely drop initial invite media, but stop dropping after re-invite

Tested by toggling re-INVITES and checking that audio was or was not heard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -804,6 +804,18 @@ Optionally included keys are:
 	Used if SDP addresses are neither trusted (through `SIP source address` or `--sip-source`) nor the
 	`media address` key is present.
 
+* `drop-traffic`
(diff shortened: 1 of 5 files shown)


rfuchs:
I think it would make more sense to have this as a dictionary key instead of an entry in the `flags`, e.g. as `drop-traffic=start` etc similar to `ICE=` or `DTLS=`. What do you think?

smititelu:
Yes, indeed. I'll update the pull request accordingly.

smititelu:
Done. Retested successfully.

PR #998, opened 2020-05-27
GH testing

mika:


mika:
Alright, so GitHub actions doesn't work *yet* here, sorry4noise.

PR #1010, opened 2020-06-05
Added "currentstatistics" to json stats.

attermann:
Added section for current server stats to easily parseable json stats.


attermann:
Re-submitted each commit as a separate pull.

Issue #1009, opened 2020-06-05
Case-sensitive codec names


attermann:
Version: 8.3.2.0+0~mr8.3.2.0 git-mr8.3-f1e8fa1
Ubuntu 16.04.6 LTS

RTPEngine is not properly handling codecs in SDP when the case does not exactly match what it expects (as specified in the offer).

I found a similar issue #775 from last year that has an associated TT#59100 and commits to resolve the issue, however either it's not clear exactly how to enable support for alternate case in codec names, or the fix is incomplete.

In my particular case, when iLBC codec is specified using the name "ILBC" instead of the expected "iLBC" as in the SDP below, RTPEngine creates a passthrough handler for "ILBC" in addition to an encoder and passes the codec through to the B-side instead of masking it.

```
v=0
o=- 8133564038 37843 IN IP4 172.26.170.170
s=ashetpb
c=IN IP4 192.168.29.227
t=0 0
a=msid-semantic:WMS 426c5657-f82d-4b34-9f8c-497964aa8428
m=audio 62428 RTP/AVP 0 8 102 101
a=rtcp-mux
a=rtpmap:101 telephone-event/8000
a=rtpmap:102 ILBC/8000
a=fmtp:101 0-15
a=fmtp:102 mode=20
a=ptime:20
a=sendrecv
a=ssrc:2282603044 cname:vDjAdJ2KvNrjcrfn
a=ssrc:2282603044 msid:426c5657-f82d-4b34-9f8c-497964aa8428 56004291-d427-4564-847f-bf4914b7ac3e
a=ssrc:2282603044 mslabel:426c5657-f82d-4b34-9f8c-497964aa8428
a=ssrc:2282603044 label:56004291-d427-4564-847f-bf4914b7ac3e
```

This results in the following RTPEngine logs
```
Jun  4 23:39:56 uc-dal rtpengine[14573]: INFO: [5363F32A0D66685D5C7FD4DD9DBAB4D4D54373AD]: Received command 'offer' from 1.2.3.4:55112
[...]

PR #1011, opened 2020-06-05
Made hash tables for flags case-insensitive.

attermann:
Made hash tables for flags case-insensitive so that CODEC names in SDP that are not the expected case will still match flags like code-mask-CODEC and codec-strip-CODEC.


attermann:
Addresses issue #1009.

PR #1013, opened 2020-06-09
Solve Vlagrind "definitely lost" memory leaks

smititelu:
1. run rtpengine under valgrind: sudo valgrind --leak-check=full --show-leak-kinds=all --track-origins=yes --verbose --log-file=valgrind-out1.txt daemon/rtpengine -f -E --no-log-timestamps --pidfile /run/ngcp-rtpengine-daemon.pid --config-file /etc/rtpengine/rtpengine1.conf
2. wait it to start
3. Ctrl+c

Vlagrind reports lots of memory warnings. Some of them are "definitely lost" (attached file) which are definitely memory leaks.

After this commit, there are 0 "definitely lost" warnings that valgrind reports(for the above steps).

[valgrind_definitely.txt](https://github.com/sipwise/rtpengine/files/4750748/valgrind_definitely.txt)

### DIFF
--- a/daemon/aux.c
+++ b/daemon/aux.c
@@ -141,8 +141,10 @@ void threads_join_all(int wait) {
 			pthread_join(*t, NULL);
 			threads_to_join = g_list_delete_link(threads_to_join, threads_to_join);
 			l = g_list_find_custom(threads_running, t, thread_equal);
-			if (l)
+			if (l) {
+				g_slice_free1(sizeof(*t), l->data);
 				threads_running = g_list_delete_link(threads_running, l);
+			}
 			else
 				abort();
 			g_slice_free1(sizeof(*t), t);
@@ -263,4 +265,12 @@ int g_tree_find_all_cmp(void *k, void *v, void *d) {
 void free_buf(char **p) {
 	if (*p)
 		free(*p);
+}
+
+void free_gbuf(char **p) {
+	g_free(*p);
+}
+
+void free_gvbuf(char ***p) {
+	g_strfreev(*p);
 }
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -68,14 +68,12 @@ struct rtpengine_config rtpe_config = {
 	.redis_allowed_errors = -1,
 	.redis_disable_time = 10,
 	.redis_connect_timeout = 1000,
-	.rec_method = "pcap",
-	.rec_format = "raw",
 	.media_num_threads = -1,
 	.dtls_rsa_key_size = 2048,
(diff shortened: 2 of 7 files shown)


rfuchs:
Thanks for this, I've had it on my own to-do list for ages...

smititelu:
Just a note that I'm currently looking also into the other "possibly lost" warnings, trying to fix as many as possible / if possible.

The best would be to solve all and have clean valgrind output. Thus, would make debugging of (new) memory leaks easier.

rfuchs:
The argument/config file handling is incomplete I think. The reason you're getting a warning about trying to free addresses in r-x mapped files is that there are defaults for these strings that are given as string literals and thus aren't allocated. But if a value for these strings were taken from a config file or argv, then they would be allocated.

Also, the 3-times iteration over arguments, then config file, then arguments again, doesn't take duplicates into account. If some values are given in both locations, the pointer would simply be overwritten without the previous value being freed (I think). Something like strdupfree() should probably be used here in some way.

The duplication of the shared arguments is also not nice. There was a reason why it was in a single place in auxlib.c, so I'd prefer to preserve that.

Moving all the option variables out of options() and into global namespace seems like the wrong solution. Those are meant to be temporary and should be freed when options() returns. Those are probably good candidates for auto-cleanup, see AUTO_CLEANUP*() macros, in particular AUTO_CLEANUP_BUF.

Let me know if you want to fiddle with this more, because I can probably address these points myself too.

smititelu:
Thanks for feedback. Yes, I'll fiddle a little more with this.

smititelu:
Updated the commit solving the following:
- add config_load_free() to free all the common options; call it in options_free() and in config_load() if any errors happen
- moved default of log prefix/suffix common options in config_load()
- add AUTO_CLEANUP_G/GVBUF and use them in options()

Related to the ```if_a``` stuff, yes i also tested with ```G_SLICE=always-malloc```, however no "definitely lost" reported. However, there are some "still reachable" warnings that may be related to that.

smititelu:
Just for the record, two TODOs left:
1. If STRING or STRING LIST param is present in both config file and cli argument valgrind reports "definitely lost" for it => in ```config_load()``` free the memory for common config file vs cli arguments, after config file was parsed but before argv is parsed second time

2. Free ```if_a``` inside ```options()``` and don't use global reference anymore. Solve everything else where if_a is currently used.

rfuchs:
I'll be working on these points now, thanks

rfuchs:
JFYI I've pushed some more cleanups to https://github.com/sipwise/rtpengine/tree/rfuchs/cleanup

They're addressing a bunch of the "still reachable" allocations

PR #1016, opened 2020-06-15
Solve Valgrind "possibly lost" for redis and dtls

smititelu:
There is 1 more warning related to this, that I am confused how to solve:
```
==30466== 64 bytes in 1 blocks are still reachable in loss record 211 of 387
==30466==    at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==30466==    by 0x522BAB8: g_malloc (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.5600.4)
==30466==    by 0x51F95A2: g_async_queue_new_full (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.5600.4)
==30466==    by 0x524EA0C: g_thread_pool_new (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.5600.4)
==30466==    by 0x12DD6F: redis_restore (redis.c:1758)
==30466==    by 0x119D82: create_everything (main.c:868)
==30466==    by 0x119D82: main (main.c:892)
```          
I think it might be related to the glibc since ```g_thread_pool_free``` should already do all the job.

I tried creating a small test program only with ```g_thread_pool_new``` ```g_thread_pool_free``` with empty thread  callback and valgrind still reported the above warning.

### DIFF
--- a/daemon/dtls.c
+++ b/daemon/dtls.c
@@ -384,6 +384,19 @@ struct dtls_cert *dtls_cert() {
 	return ret;
 }
 
+void dtls_cert_free(void) {
+	rwlock_lock_w(&__dtls_cert_lock);
+
+	if (__dtls_cert)
+		obj_put(__dtls_cert);
+
+	__dtls_cert = NULL;
+
+	rwlock_unlock_w(&__dtls_cert_lock);
+
+	return ;
+}
+
 static int verify_callback(int ok, X509_STORE_CTX *store) {
 	SSL *ssl;
 	struct dtls_connection *d;
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -942,6 +942,11 @@ int main(int argc, char **argv) {
 
 	ilog(LOG_INFO, "Version %s shutting down", RTPENGINE_VERSION);
 
+	redis_close(rtpe_redis);
+	redis_close(rtpe_redis_write);
+	redis_close(rtpe_redis_notify);
+	dtls_cert_free();
+
 	unfill_initial_rtpe_cfg(&initial_rtpe_config);
 
 	options_free();
--- a/daemon/redis.c
+++ b/daemon/redis.c
@@ -631,6 +631,9 @@ void redis_notify_loop(void *d) {
 		}
 	}
 
+	// free libevent
+	libevent_global_shutdown();
+
 	// unsubscribe notifications
 	redis_notify_subscribe_action(UNSUBSCRIBE_ALL, 0);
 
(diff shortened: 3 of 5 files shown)


smititelu:
The test scenario is the same as in #1013

rfuchs:
What you're seeing is probably the `unused_thread_queue` from https://github.com/GNOME/glib/blob/mainline/glib/gthreadpool.c#L111

It's created when the first thread pool is created: https://github.com/GNOME/glib/blob/mainline/glib/gthreadpool.c#L581

I don't see any code trying to free it ever (should be `g_async_queue_unref`) so I think this is to be expected.

daemon/dtls.c
```diff
@@ -384,6 +384,14 @@ struct dtls_cert *dtls_cert() {
 	return ret;
 }
 
+void dtls_cert_free(void) {
+	rwlock_lock_r(&__dtls_cert_lock);
+	obj_put(__dtls_cert);
+	rwlock_unlock_r(&__dtls_cert_lock);
```
rfuchs, line 390: This should probably be a write lock, together with setting `__dtls_cert = NULL`, but since we're in the shutdown here anyway, it doesn't matter much.

rfuchs:
Looks like I'm getting a shutdown crash with this patch applied:
```
[pid 3177782] madvise(0x7f965f0c2000, 8368128, MADV_DONTNEED) = 0
[pid 3177782] exit(0)                   = ?
[pid 3177782] +++ exited with 0 +++
<... futex resumed>)                    = 0
munmap(0x7f9616ffe000, 8392704)         = 0
futex(0x564974b81e80, FUTEX_WAKE_PRIVATE, 1) = 0
write(2, "[1592230466.693114] INFO: Version 8.5.0.0+0~mr8.5.0.0 git-master-a0895fa7 shutting down\n", 88) = 88
--- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_MAPERR, si_addr=0x68} ---
+++ killed by SIGSEGV (core dumped) +++
```
```
Reading symbols from ../daemon/rtpengine...
[New LWP 3177779]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `../daemon/rtpengine --config-file=none -t -1 -i 203.0.113.1 -i 2001:db8:4321::1'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000564974b10ac4 in redis_close (r=0x0) at redis.c:684

warning: Source file is more recent than executable.
684	void redis_close(struct redis *r) {
(gdb) bt
#0  0x0000564974b10ac4 in redis_close (r=0x0) at redis.c:684
#1  0x0000564974afc71a in main (argc=<optimized out>, argv=<optimized out>) at main.c:952
(gdb) 
```

smititelu:
Added null-checks before free-ing dtls and redis.

Created a new GNOME/glib issue for the above ```g_thread_pool_new```/```g_thread_pool_free```.

PR #1020, opened 2020-06-17
Few more valgrind fixes

smititelu:
- free the allocate str_init_dup() for if name
- clear the queues for keyspaces and interfaces config

### DIFF
--- a/lib/auxlib.c
+++ b/lib/auxlib.c
@@ -102,16 +102,20 @@ void config_load_free(struct rtpengine_common_config *cconfig) {
 }
 
 static void free_gkeyfile(GKeyFile **k) {
-	g_key_file_free(*k);
+	if (k && *k)
+		g_key_file_free(*k);
 }
 static void free_gopte(GOptionEntry **k) {
-	free(*k);
+	if (k && *k)
+		free(*k);
 }
 static void free_goptc(GOptionContext **k) {
-	g_option_context_free(*k);
+	if (k && *k)
+		g_option_context_free(*k);
 }
 static void free_gerror(GError **k) {
-	g_error_free(*k);
+	if (k && *k)
+		g_error_free(*k);
 }
 
 void config_load(int *argc, char ***argv, GOptionEntry *app_entries, const char *description,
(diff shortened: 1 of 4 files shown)


lib/auxlib.c
```diff
@@ -102,16 +102,20 @@ void config_load_free(struct rtpengine_common_config *cconfig) {
 }
 
 static void free_gkeyfile(GKeyFile **k) {
-	g_key_file_free(*k);
+	if (k && *k)
+		g_key_file_free(*k);
```
rfuchs, line 106: These shouldn't be necessary as the pointer in an auto-cleanup function should always be non-null. But doesn't hurt either of course.

rfuchs:
Getting a shutdown crash with this.
```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f0ba1ffa55b in __GI_abort () at abort.c:79
#2  0x00007f0ba2053038 in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7f0ba215ff3e "%s\n")
    at ../sysdeps/posix/libc_fatal.c:181
#3  0x00007f0ba205a3da in malloc_printerr (str=str@entry=0x7f0ba215e0e0 "free(): invalid pointer") at malloc.c:5339
#4  0x00007f0ba205bdcc in _int_free (av=<optimized out>, p=<optimized out>, have_lock=0) at malloc.c:4173
#5  0x00007f0ba4b0ff2f in g_queue_foreach () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#6  0x00007f0ba4b0ffb5 in g_queue_clear_full () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#7  0x00005648f448c7fd in options_free () at main.c:975
#8  main (argc=<optimized out>, argv=<optimized out>) at main.c:975
```
I believe it's because `if_addr_parse` uses `g_slice_alloc` to populate the list of interfaces in `rtpe_config.interfaces`, while your added code uses malloc/free to duplicate and free the `ini_rtpe_cfg->interfaces` list.

smititelu:
Thanks for pointing this out. I missed it since I was testing only with ```G_SLICE=always-malloc```.

Updated the commit so both will use g_slice_alloc/free. Added also ```str_free_dup()``` function.

Issue #1021, opened 2020-06-19
CRIT: Fatal error: Bad command line: Key file does not have group “ ��”


amessina:
Running master@203f502006af94dad51056820bc9ae63cac69cca with
```
ExecStart=/usr/sbin/rtpengine -f --config-file=/etc/rtpengine/rtpengine.conf --config-section=rtpengine --pidfile=/run/rtpengine/rtpengine.pid
```
yelds
```
CRIT: Fatal error: Bad command line: Key file does not have group “ ��”
```
Quick review of recent commits shows a maybe related to aa98d7f86c3cee11fbc24f7340e8223dda73564a

My `rtpengine.conf` does specify `[rtpengine]` as a section.  When I remove `--config-section` from the above command line, RTPengine starts without issue.

smititelu:
I'll have a look.

smititelu:
Please check the commit in #1022 and confirm it solves the issue.

amessina:
This does fix the error, though the parsing updates seem to indicate other issues:

**/etc/rtpengine/rtpengine.conf**
```[rtpengine]
table = 0
interface=pub/<EXTERNAL_IPV4>;pub/<IPV6>;priv/<INTERNAL_IPV4>;priv/<IPV6>
iptables-chain = RTPENGINE_allow
listen-cli=127.0.0.1:9900
listen-ng = 127.0.0.1:12221
tos = 184
recording-dir = /var/spool/rtpengine
recording-method = proc
log-level = 6
```

```
WARNING: [0_426106263@2001:0db8:85a3:0000:0000:8a2e:0370:7334]: Interface 'priv' not found, using default
WARNING: [0_426106263@2001:0db8:85a3:0000:0000:8a2e:0370:7334]: Interface 'priv' not found, using default
```


```
INFO: [0_426106263@2001:0db8:85a3:0000:0000:8a2e:0370:7334]: Received command 'offer' from 127.0.0.1:57642
DEBUG: [0_426106263@2001:0db8:85a3:0000:0000:8a2e:0370:7334]: Dump for 'offer' from 127.0.0.1:57642: { "supports": [ "load limit" ], "sdp": "v=0
o=- 20427 20427 IN IP6 2001:0db8:85a3:0000:0000:8a2e:0370:7334
s=SDP data
c=IN IP6 2001:0db8:85a3:0000:0000:8a2e:0370:7334
t=0 0
m=audio 11810 RTP/SAVP 9 18 0 101
[...]

PR #1022, opened 2020-06-19
Fix config use_section

smititelu:


smititelu:
Fixes #1021

rfuchs:
My mistake I think. Thanks.

PR #1026, opened 2020-06-24
Fix a few more "still reachable". End all redis unused threads.

smititelu:
I'm starting to test valgrind rtpengine in production environment with production config.

Found a few more "still reachable" upon start-stop.


smititelu:
Added some free in recording-daemon/main.c

Issue #1032, opened 2020-07-01
Video RTP packets erroneously briefly forwarded to audio port


jovermann:
Setup: rtpengine acts as RTP proxy between an external UA providing video (camera and no display) and an internal UA (display and no camera). rtpengine driven by Kamailio 5.2.1 using the ng protocol. Happens with rtpengine mr8.3.1 and master (yesterday).

This happens:
The external UA sends an invite with audio+video in the SDP, the internal UA responds immediately with Session Progress, requesting video only in the SDP. The internal UA display the video as expected. Eventually the user will accept the call on the internal UA with 200 OK, requesting audio+video in the SDP. Immediately after the internal UA picks up the call (200 OK), rtpengine forwards about 20-40 video frames to the audio port. After that all packets go to the correct ports.

(Background: Why is this a problem?
A sane internal UA would simply drop the video packets received on the audio, or even process them using the video decoder. The problem was only noticed since the audio receiver in the internal UA chokes on the video packets and stops processing audio, leading to "no audio from external to internal", although rtopengine continues to forward all audio packets correctly to the internal UA at any point in time. This is a bug in the internal UA which is out of scope of this issue, just to explain why the problem in rtpengine was detected. Fixing the internal UA would also fix the "no audio" problem, but this is not always possible.)

Analysis: The 20-40 video packets going to the wrong port are the packets between the point in time when rtpengine (driven by Kamailio) already switched from video-only to audio-video, and the point when the external UA processes the 200 OK SDP generated by rtpengine and switches to the new video port.

All SDP messages looks good and all UAs send to the expected ports at any time. Just rtpengine

This is what happens concretely in terms of ports:
1. Session Progress: rtpengine asks video to be sent to port 31848
2. External UA startes sending to 31848. OK.
3. Internal UA answers call. Rtpengine asks  for audio on port 31878 and video on port 31862. OK. All differnt ports. No confusion.
4. External UA continues to send video packets to 31848 until it processed the 200 OK. Still OK.
5. Rtpengine erroneously sends the video packets received on the video port 31848 to the audio port of the internal UA. Error.
6. Eventually the external UA switches sending video packets from port 31848 to the new video port 31862. OK.
7. Rtpengine forwards all packets correctly from now on.

Theory: Apparently rtpengine re-uses media slots (the things called "Media #1 and Media #2 in the INFO log messages) from video to audio when switching from video-only to audio+video. It does not firmly associate the receiving and sending ports with the media streams transported over them.

[...]

PR #1033, opened 2020-07-01
do not forward RTP packets with unsuitable payload

jovermann:
Fixing #1032

This is a last-resort check to stop RTP packets being sent to sink
ports for payloads which are not supported by the sink ports.

### DIFF
--- a/daemon/media_player.c
+++ b/daemon/media_player.c
@@ -227,6 +227,25 @@ static void __send_timer_send_common(struct send_timer *st, struct codec_packet 
 				FMT_M(sockaddr_print_buf(&st->sink->endpoint.address),
 				st->sink->endpoint.port));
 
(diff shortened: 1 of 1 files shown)


jovermann:
I consider this a workaround for a problem created deeper inside rtpengine (routing packets to wrong sinks), but I struggle to find where to properly fix this. Any comments and better solutions welcome.

Even with a better fix in place, this patch may be useful as a last-resort check to prevent packets going the wrong way.

rfuchs:
While I appreciate the effort, I don't think this is an appropriate change to make. Rtpengine is supposed to forward packets even if it doesn't understand them at all, e.g. packets that aren't even RTP, or an unknown protocol, or an unknown codec. Only packets that it can understand should be consumed or blocked, not the other way around.

PR #1049, opened 2020-07-23
Added message compression/uncompression

pchero:
Currently, the RTPEngine sends the UDP message without any. compression.

This is fine for most of the cases, but with the complex SDP, the RTPEngine generated over 3000 characters of RTP stats at the end of the call.

Added option for compression the message if the message size is over a certain size.


pchero:
Sorry, this is the wrong Pull request. I will be back with a proper request later.

PR #1052, opened 2020-07-27
Added compress option

pchero:
Added the compress/uncompress option when the RTPEngine is sending/receiving the Bencode message.

### DIFF
--- a/daemon/compress.c
+++ b/daemon/compress.c
@@ -1,0 +1,74 @@
+
+#include <str.h>
+#include <zlib.h>
+#include <string.h>
+#include <errno.h>
+
+#include "log.h"
+#include "udp_listener.h"
+#include "str.h"
+#include "compress.h"
+
+#define RTPENGINE_ZLIB_WINDOWBITS 15
+#define RTPENGINE_ZLIB_ENABLE_ZLIB_GZIP 32
+
+static void init_stream(z_stream *stream)
+{
+	stream->zalloc = Z_NULL;
+	stream->zfree = Z_NULL;
+	stream->opaque = Z_NULL;
+}
+
+// compress_data compress the data of the given length.
+// returns compressed buffer length and replace the original buffer.
+int compress_data(char *buf, int len, char *out_buf, int out_len) {
+	uLong comp_len = compressBound(len);
+	int ret;
+
+	if (out_len < comp_len) {
+		ilog(LOG_ERROR,"The given outbuf is not enough for compressed message\n");
+		return -1;
+	}
+
+	ret = compress2((Bytef *)out_buf, &comp_len, (Bytef *)buf, len, Z_BEST_COMPRESSION);
+	if (ret != Z_OK) {
+		ilog(LOG_ERROR,"Could not compress the message. err: %d\n", ret);
+		return -1;
+	}
+
+	return comp_len;
+}
+
+// uncompress_data uncompress the compressed data with given length.
+// it compare the first byte to check the compressed data or not.
+int uncompress_data(char *buf, int len, char *buf_tmp, int buf_size) {
+	z_stream stream;
+	int ret;
+
+	init_stream(&stream);
+	ret = inflateInit2(&stream, RTPENGINE_ZLIB_WINDOWBITS | RTPENGINE_ZLIB_ENABLE_ZLIB_GZIP);
+	if (ret != Z_OK) {
+		ilog(LOG_ERROR,"Could not initiate zstream. err: %d\n", ret);
+		return -1;
+	}
+
+	stream.next_in = (Bytef *)buf;
+	stream.avail_in = len;
+
+	memset(buf_tmp, 0x00, buf_size);
+	stream.next_out = (Bytef *)buf_tmp;
+	stream.avail_out = buf_size - 1;
+
+	ret = inflate(&stream, Z_NO_FLUSH);
+	if (ret != Z_OK && ret != Z_STREAM_END) {
+		inflateEnd(&stream);
+		ilog(LOG_ERROR,"Could not uncompress the data correctly. err: %d\n", ret);
+		return -1;
+	}
+
+	memcpy(buf, buf_tmp, buf_size);
+	ret = stream.total_out;
+	inflateEnd(&stream);
+
+	return ret;
+}
--- a/daemon/control_ng.c
+++ b/daemon/control_ng.c
@@ -16,6 +16,7 @@
 #include "log_funcs.h"
 #include "main.h"
 #include "statistics.h"
+#include "compress.h"
 
 
 mutex_t rtpe_cngs_lock;
@@ -123,8 +124,10 @@ static void control_ng_incoming(struct obj *obj, str *buf, const endpoint_t *sin
 	bencode_item_t *dict, *resp;
 	str cmd = STR_NULL, cookie, data, reply, *to_send, callid;
 	const char *errstr, *resultstr;
-	struct iovec iov[3];
-	unsigned int iovlen;
+	char send_buf[0x10000];
+	char comp_buf[0x10000];
+	char *send_target;
+	int total_len;
 	GString *log_str;
 	struct timeval cmd_start, cmd_stop, cmd_process_time;
 	struct control_ng_stats* cur = get_control_ng_stats(c,&sin->address);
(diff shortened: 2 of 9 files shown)


guillemj:
Hi!

I'm going to perform a cursory review. But Richard is the one who will eventually approve this.

First off, please squash the various commit fixups into logic commits, in this case I think one commit would do, and then force push. Also it looks like the prevalent coding style is to use hard-tabs instead of spaces, please fix those too.

daemon/compress.c
```diff
@@ -0,0 +1,75 @@
...
+#include <errno.h>
+
+#include "log.h"
+#include "udp_listener.h"
+#include "str.h"
+#include "compress.h"
+
+#define windowBits 15
+
+#define GZIP_ENCODING 16
```
guillemj, line 14: This looks unued.

daemon/compress.c
```diff
@@ -0,0 +1,75 @@
...
+#define ENABLE_ZLIB_GZIP 32
+
+static void init_stream(z_stream *stream);
```
guillemj, line 17: This forward declaration is unnecessary, as we are just defining it in the next line.

daemon/compress.c
```diff
@@ -0,0 +1,75 @@
...
+#include <zlib.h>
+#include <string.h>
+#include <errno.h>
+
+#include "log.h"
+#include "udp_listener.h"
+#include "str.h"
+#include "compress.h"
+
+#define windowBits 15
```
guillemj, line 12: While this is defined after the zlib.h inclusion, it matches the name of the function argument. I'd use something more descriptive and less generic, perhaps `#define RTPENGINE_ZLIB_WINDOWBITS 15`?

daemon/compress.c
```diff
@@ -0,0 +1,75 @@
...
+
+#define GZIP_ENCODING 16
+#define ENABLE_ZLIB_GZIP 32
```
guillemj, line 15: Perhaps rename this too, to something that matches the window bits macro?

daemon/compress.c
```diff
@@ -0,0 +1,75 @@
...
+	stream->zfree = Z_NULL;
+	stream->opaque = Z_NULL;
+}
+
+
+// compress_data compress the data of the given length.
+// returns compressed buffer length and replace the original buffer.
+int compress_data(char *buf, int len) {
+	uLong comp_len = compressBound(len);
+    char tmp[comp_len];
```
guillemj, line 30: I'm not sure what's @rfuchs take on VLAs?

daemon/compress.c
```diff
@@ -0,0 +1,75 @@
...
+{
+	stream->zalloc = Z_NULL;
+	stream->zfree = Z_NULL;
+	stream->opaque = Z_NULL;
+}
+
+
+// compress_data compress the data of the given length.
+// returns compressed buffer length and replace the original buffer.
+int compress_data(char *buf, int len) {
```
guillemj, line 28: I don't think it's of very good form to use the same buffer for input and output, because we end up needing a temporary that we then need to copy, which looks inefficient.

daemon/compress.c
```diff
@@ -0,0 +1,75 @@
...
+	uLong comp_len = compressBound(len);
+    char tmp[comp_len];
+    int ret;
+
+    ret = compress2((Bytef *)tmp, &comp_len, (Bytef *)buf, len, Z_BEST_COMPRESSION);
+    if (ret != Z_OK) {
+        return -1;
```
guillemj, line 35: Log the error, as it is done also on decompression?

daemon/compress.c
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    memcpy(buf, tmp, comp_len);
+    return comp_len;
+}
+
+// uncompress_data uncompress the compressed data with given length.
+// it compare the first byte to check the compressed data or not.
+int uncompress_data(char *buf, int len, char *buf_tmp, int buf_size) {
```
guillemj, line 44: Same with the temporary buffer here, I'd just use different in and out buffers.

daemon/control_ng.c
```diff
@@ -335,16 +336,15 @@ static void control_ng_incoming(struct obj *obj, str *buf, const endpoint_t *sin
...
 
-	iov[0].iov_base = cookie.s;
-	iov[0].iov_len = cookie.len;
-	iov[1].iov_base = " ";
-	iov[1].iov_len = 1;
-	iov[2].iov_base = to_send->s;
-	iov[2].iov_len = to_send->len;
-
-	socket_sendiov(ul, iov, iovlen, sin);
+	total_len = snprintf(send_buf, sizeof(send_buf), "%.*s %.*s", cookie.len, cookie.s, to_send->len, to_send->s);
```
guillemj, line 340: I'd assume the `snprintf()` switch here is going to be substantially slower than the iov send, But perhaps this is not performance critical? (I don't know). Maybe if cookie is not big, then only compressing to_send would be better, and then we just need to set `iov[2]` to the original or compressed buffer depending on whether we are compressing?

daemon/compress.c
pchero: If you are saying so, I'd fix it.

But to return the compressed data it required allocating the memory using malloc()/calloc, and need to be freed them afterward. I thought that one is more expensive.

Before fixing it, I would like to listen to your opinion again.

daemon/compress.c
pchero: What is the VLAs?

daemon/compress.c
pchero: Yes, that might be better.
But if you see where this function is used, you can understand it.

This uncompress_Data is being used in the infinite loop at the udp_listener_incoming(). So, rather than use the different buffer and make them allocating/deallocating the memory all the time for each message receiving, add the args for buffer and reuse them all the time is better.

daemon/compress.c
pchero: Agree, removed.

daemon/compress.c
guillemj: A VLA is a variable-length array.

daemon/compress.c
guillemj: I don't think there's a need for memory allocation here, I think it would just be a matter of moving this around. You can just use another buffer in the call site, and pass it to the function. If we are doing compression then you can simply assign the output buffer to be passed to the send function, otherwise we pass the input one. This avoid allocation and an additional buffer copy, as we only need to copy a pointer instead.

daemon/control_ng.c
guillemj: I guess it depends on what can appear in the uncompressed to_send buffer. Otherwise it would be a matter of a conditional I think? Something like the following pseudo-code:

```
  setup iov[0] from cookie
  setup iov[1] from " "

  if (compress(in, out) == OK) {
    setup iov[2] from out
  } else {
    setup iov[2] from in
  }
```

of course that means the receiving end needs to handle the decompression after parsing the message instead of just on  ingestion. I'll leave @rfuchs to decide on this.

daemon/compress.c
pchero: Np, I've got that. Fixed it. 👍

daemon/compress.c
pchero: I still have the same idea, but I'm OK to fix it if you want to. :)

daemon/compress.c
pchero: Thanks, now I've got that. Btw, it's gone now. :) I'm resolving this.

daemon/control_ng.c
pchero: I don't have that strong opinion about this. But I was thinking doing like this will make it easier to do handle(uncompress) the message for the other third party.(because this one was easier to me when I fix it the Kamailio's RTPEngine module).

I'm just waiting for the decision.

include/compress.h
```diff
@@ -0,0 +1,25 @@
+#ifndef __CODEC_H__
```
rfuchs, line 1: This should be __COMPRESS_H__ not __CODEC_H__

daemon/udp_listener.c
```diff
@@ -47,6 +50,13 @@ static void udp_listener_incoming(int fd, void *p, uintptr_t x) {
 			return;
 		}
 
+		if (is_compressed(buf)) {
+			ret = uncompress_data(buf, len, buf_tmp, 0x10000);
```
rfuchs, line 54: This irks me for two reasons.

1) Using a single leading lower case `x` to detect compressed data seems very unreliable, and
2) The `udp_listener` facility is not exclusive to the NG protocol support.

Those two together make this approach look very prone to false positives. Instead I would suggest to either use a stronger magic cookie to mark compressed data, OR move the support for decompression into `control_ng`. Or maybe both :smile:

daemon/control_ng.c
rfuchs: I haven't looked at the Kamailio portion of this yet, but generally I would agree that this extra `snprintf` could/should be avoided. You could leave the message format as it is (cookie plus space plus compressed payload), or even use the bencode facility to carry the compressed data, e.g. the message would still be cookie plus space plus bencode dictionary, only that the dictionary would be something like `{ 'zlib': '... raw data ...' }` with the compressed data then decompressing to the actual bencode dictionary. The latter could be easily implemented by extended the iovec array by a few fields and hardcoding its format thanks to the simplicity of bencode. This would also address my concerns about `is_compressed()` as well as saving the daemon from having to uncompress first in order to detect and handle retransmits.

include/compress.h
pchero: 😱

pchero:
Hi @guillemj , @rfuchs ,

Thank you for your kind review. But I don't think this change is valid and useful anymore, at this moment.
So, I'm closing this MR now.

Thank you for all messages.

Kind regards,
Sungtae

rfuchs:
@pchero If this was supposed to address UDP sending issues related to large packet sizes, MTU, and broken packet fragmentation, we have a feature in the works that adds HTTP and Websocket transports, so it should be possible to use those for rtpengine signalling in the future.

pchero:
Thank you @rfuchs !! 👍 👍 👍

PR #1072, opened 2020-09-07
Add redis async delete

smititelu:
Add a new thread to handle async deletes, using libevent, when enabled by config; default is disabled.

Note that when redis disconnects, deletes are currently lost.

@rfuchs let me know what you think of this PR.

### DIFF
--- a/daemon/cli.c
+++ b/daemon/cli.c
@@ -995,7 +995,7 @@ static void cli_incoming_ksadd(str *instr, struct streambuf *replybuffer) {
 		rwlock_lock_w(&rtpe_config.config_lock);
 		if (!g_queue_find(&rtpe_config.redis_subscribed_keyspaces, GUINT_TO_POINTER(uint_keyspace_db))) {
 			g_queue_push_tail(&rtpe_config.redis_subscribed_keyspaces, GUINT_TO_POINTER(uint_keyspace_db));
-			redis_notify_subscribe_action(SUBSCRIBE_KEYSPACE, uint_keyspace_db);
+			redis_notify_subscribe_action(rtpe_redis_notify, SUBSCRIBE_KEYSPACE, uint_keyspace_db);
 			streambuf_printf(replybuffer, "Success adding keyspace %lu to redis notifications.\n", uint_keyspace_db);
 		} else {
 			streambuf_printf(replybuffer, "Keyspace %lu is already among redis notifications.\n", uint_keyspace_db);
@@ -1024,7 +1024,7 @@ static void cli_incoming_ksrm(str *instr, struct streambuf *replybuffer) {
                 streambuf_printf(replybuffer, "Fail removing keyspace %s to redis notifications; no digists found\n", instr->s);
 	} else if ((l = g_queue_find(&rtpe_config.redis_subscribed_keyspaces, GUINT_TO_POINTER(uint_keyspace_db)))) {
 		// remove this keyspace
-		redis_notify_subscribe_action(UNSUBSCRIBE_KEYSPACE, uint_keyspace_db);
+		redis_notify_subscribe_action(rtpe_redis_notify, UNSUBSCRIBE_KEYSPACE, uint_keyspace_db);
 		g_queue_remove(&rtpe_config.redis_subscribed_keyspaces, l->data);
 		streambuf_printf(replybuffer, "Successfully unsubscribed from keyspace %lu.\n", uint_keyspace_db);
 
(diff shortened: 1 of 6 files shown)


rfuchs:
Is this because you find the `DEL` command takes too long to process?

smititelu:
Yes, actually we have a setup where the redis write server is in another database and a ICMP ping to it takes long.

Overall, for our setup, a ```DEL``` command takes even longer, to send and consume.

Updated the commit such as the new delete async thread to just pool the redis write state, upon redis disconnection, and don't do anything to the redis write context.

smititelu:
Update: I'm having a hard time synchronizing rtpe_redis_delete_async_context between the delete async thread and other rtpengine threads, under production-like traffic. This is because: ```The```(async)```context object is always free'd after the disconnect callback fired.``` This leads to a segfault in redis_delete_async_call_json()->redisAsyncCommand().

I will try a different approach:
1. all other threads push redis delete commands strings into a glib queue, under a lock
2. only delete_async thread will handle redis async ctx; will pop redis delete commands strings for the glib queue, under a lock, and send them, using libevent
3. Need to periodically send loopbreak to delete_async thread in order for step 2 to happen

What do you think of this different approach?

rfuchs:
`redis_delete_async_call_json()` uses `rtpe_redis_delete_async_context` under a lock and only if non-null. `redis_delete_async_context_disconnect()` sets `rtpe_redis_delete_async_context` to null, also under a lock, and the implicit `free` only happens after the callback is done. I'm not sure how a segfault can happen here? If you see a segfault then there might be a different underlying issue.

One thing I notice is that the `struct redis` (`rtpe_redis_write`) seems to be shared between the async connection and the regular non-async one, which means they (erroneously?) share the connection state, even without being under a lock (in `redis_delete_async_loop()`). You probably need to keep track of the connection state separately, or even better, refactor `struct redis` to be able to handle both async and non-async Redis connections.

In general I see a lot of code duplication between the notification code and the new async delete code. Can that be refactored so they share most of the code? Do we really need two event bases? (I guess you need two if you want to strictly separate the threads and what they do.)

If you want to proceed with separating handling the connection into a separate thread, glib provides some convenience structures for that, namely `GAsyncQueue` or even (somewhat different) `GThreadPool`. But I feel that it shouldn't be necessary and might just be working around some other issue.

smititelu:
I updated the code:
- remove code duplication
- add some async params to struct redis
- the delete_async thread will be the only one that handles redis write async_ctx
- use async_queue to decouple "delete" commands from delete_async thread vs other rtpengine call threads
- use async_mutex to guard async_queue
- use timer poller loop to trigger delete_async loopbreak, depending on newly added config param redis-delete-async-interval

Preliminary pre-production testing looks ok. Will test this a little more and update the status.

smititelu:
Added a small fix related to the locking of the async queue, in redis_delete_async() and parameters rename.

Testing the feature since last comment, no more issues spotted.

rfuchs:
This segfaults on startup for me.

```
Core was generated by `./rtpengine --config-file=none -t -1 -i 203.0.113.1 -i 2001:db8:4321::1 -n 2223'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00005574b45b89c9 in poller_timer_loop (d=0x5574b5eff120) at poller.c:529
529			if (rtpe_redis_write->async_last + rtpe_config.redis_delete_async_interval <= rtpe_now.tv_sec) {
[Current thread is 1 (Thread 0x7f1280237700 (LWP 117351))]
(gdb) bt
#0  0x00005574b45b89c9 in poller_timer_loop (d=0x5574b5eff120) at poller.c:529
#1  0x00005574b45b8bf2 in thread_detach_func (d=0x5574b5ee12e0) at aux.c:209
#2  0x00007f12899ecea7 in start_thread (arg=<optimized out>) at pthread_create.c:477
#3  0x00007f128991ceaf in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb) bt full
#0  0x00005574b45b89c9 in poller_timer_loop (d=0x5574b5eff120) at poller.c:529
        tv = {
          tv_sec = 1601384704,
          tv_usec = 518507
        }
        wt = <optimized out>
#1  0x00005574b45b8bf2 in thread_detach_func (d=0x5574b5ee12e0) at aux.c:209
        dt = 0x5574b5ee12e0
        t = 0x7f1278001bb0
        scheduler = <optimized out>
#2  0x00007f12899ecea7 in start_thread (arg=<optimized out>) at pthread_create.c:477
        ret = <optimized out>
        pd = <optimized out>
        unwind_buf = {
          cancel_jmp_buf = {{
              jmp_buf = {139717435946752, 6713012582790149102, 140730246019134, 140730246019135, 139717435941184, 8396800, -6697862932787678226, -6697844005462002706},
              mask_was_saved = 0
            }},
          priv = {
            pad = {0x0, 0x0, 0x0, 0x0},
            data = {
              prev = 0x0,
              cleanup = 0x0,
              canceltype = 0
            }
          }
        }
        not_first_call = 0
#3  0x00007f128991ceaf in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

rfuchs:
Also can you squash those commits into one while you're at it please :)

smititelu:
Fixed the above segfault at startup and squashed commits.

PR #1073, opened 2020-09-07
EL packaging: Build with various FreeSWITCH package names for spandsp

space88man:
FreeSWITCH >= 1.10.5 has two packages names: spandsp or spandsp3.

As of 2020-09-08,  FreeSWITCH provides for EL systems:
- spandsp-devel-1.99.0-15.a6266f2259.x86_64.rpm
- spandsp3-devel-3.0.0-25.6ec23e5a7e.x86_64.rpm

They both provide `pkgconfig(spandsp)`, but unfortunately, spandsp3-devel does not provide `spandsp-devel` to act as a substitute.

The package from EPEL also provides `pkgconfig(spandsp)`, so this PR shouldn't break existing builds.

### DIFF
--- a/el/rtpengine.spec
+++ b/el/rtpengine.spec
@@ -17,7 +17,7 @@ BuildRequires:	glib2-devel libcurl-devel openssl-devel pcre-devel
 BuildRequires:	xmlrpc-c-devel zlib-devel hiredis-devel
 BuildRequires:	libpcap-devel libevent-devel json-glib-devel 
 BuildRequires:	gperf perl-IPC-Cmd
-BuildRequires:  spandsp-devel
+BuildRequires:  pkgconfig(spandsp)
 Requires(pre):	shadow-utils
 
 %if 0%{?with_transcoding} > 0


space88man:
If accepted, any chance this could be back-ported to  `mr9.0` as well since that is at early stage of branching?

rfuchs:
Do you happen to know how backwards compatible this is? Could it cause problems when building for older EL releases?

space88man:
This should work back to EL6. I tested on EL7 and checking the EL6 RPM from EPEL:

```
$ rpm -qp --provides spandsp-devel-0.0.6-0.8.pre21.el6.x86_64.rpm 
warning: spandsp-devel-0.0.6-0.8.pre21.el6.x86_64.rpm: Header V3 RSA/SHA256 Signature, key ID 0608b895: NOKEY
pkgconfig(spandsp) = 0.0.6
spandsp-devel = 0.0.6-0.8.pre21.el6
spandsp-devel(x86-64) = 0.0.6-0.8.pre21.el6
```

Also EL6 is EOL on 11/2020 so maybe that should be enough?

PR #1082, opened 2020-09-25
fix segfault upon got reINVITE and BYE at the same time

sem32:


daemon/call.c
```diff
@@ -1549,7 +1549,10 @@ static void __sdes_accept(struct call_media *media, const struct sdp_ng_flags *f
 		}
 	}
 
-	struct crypto_params_sdes *cps_in = media->sdes_in.head->data;
+	struct crypto_params_sdes *cps_in = media->sdes_in.head ? media->sdes_in.head->data : NULL;
+	if (cps_in == NULL) {
+		return;
+	}
```
guillemj, line 1555: I'd probably simplify this to check whether sdes_in.head is NULL and then immediately return (also no need for {}), instead of checking twice first with the ternary operator and then the if. Or is this to also cover for .data possibly being NULL (if that's possible at all)?

(In any case @rfuchs is the one who knows the codebase, and will chime in.)

guillemj:
Ah perfect then thanks. Could you, though, squash both commits into a single one, so that the history does not get confusing?

PR #1083, opened 2020-09-28
fix segfault upon got reINVITE and BYE at the same time

sem32:


rfuchs:
I think there's a different underlying problem here. The function starts with
```
	if (!media->sdes_in.length)
		return;
```
So `media->sdes_in.head == NULL` later on should never happen. The only way this can happen is due to a missing lock somewhere, and a race condition. Can you explain the circumstances in which you see this segfault?

rfuchs:
Ah I guess you're using a number of `SDES-no-...` options?

sem32:
Yes, you are right, we are using 
```trust-address replace-origin replace-session-connection strict-source ICE=remove RTP/SAVP SDES-encrypted_srtp DTLS=off SDES-no-AES_CM_256_HMAC_SHA1_32 SDES-no-AES_CM_192_HMAC_SHA1_32 SDES-no-AES_CM_192_HMAC_SHA1_80 SDES-no-AES_CM_128_HMAC_SHA1_32 SDES-no-F8_128_HMAC_SHA1_80 SDES-no-F8_128_HMAC_SHA1_32 SDES-no-NULL_HMAC_SHA1_80 SDES-no-NULL_HMAC_SHA1_32 SDES-no-AES_256_CM_HMAC_SHA1_32 SDES-no-AES_256_CM_HMAC_SHA1_80 SDES-no-AES_192_CM_HMAC_SHA1_32 SDES-no-AES_192_CM_HMAC_SHA1_80 rtcp-mux-demux direction=eth0 direction=eth0 via-branch=extra```

rfuchs:
Ok that explains then, so there just weren't any cipher suites left :)

PR #1099, opened 2020-10-28
Add format_cmp_f stub when transcoding is not enabled

smititelu:
needed in order to compile when transcoding is not enabled


rfuchs:
I should really add that compilation option to our tests...

Issue #793, opened 2019-06-11
no rtcp-mux attribute in answer response for multiple branches


korayvt:
Hi,

I attached two different log files.

One is for single branch call (rctp-mux.log) and the other is for multiple branch call (no-rtcp-mux.log). 
Although the command for the answer is similar in those two scenarios, I could not get rtcp-mux attribute in answer response sdp in multiple branch call.

Am I missing something?

[no-rtcp-mux.log](https://github.com/sipwise/rtpengine/files/3274989/no-rtcp-mux.log)
[rtcp-mux.log](https://github.com/sipwise/rtpengine/files/3274990/rtcp-mux.log)

rfuchs:
Can confirm that this is not working as expected

PR #1126, opened 2020-12-04
Update call.c

jakubkarolczyk:
If the consecutive answer is from different media endpoint (e.g. early media scenario with different media endpoint used for 183), then we are loosing RTCP_MUX_OVERRIDE, as it was set for the previous answer. We need to explicitly check if rtcp-mux was initially offered by the client.

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -1653,6 +1653,11 @@ static void __rtcp_mux_logic(const struct sdp_ng_flags *flags, struct call_media
 	if (flags->opmode == OP_ANSWER) {
 		/* default is to go with the client's choice, unless we were instructed not
 		 * to do that in the offer (see below) */
+		
+		/* if the consecutive answer is from different media endpoint, then we need to check if rtcp-mux was offered initially by client */
+                if (MEDIA_ISSET(media, RTCP_MUX))
+                    return;
+		
 		if (!MEDIA_ISSET(other_media, RTCP_MUX_OVERRIDE))
 			bf_copy_same(&media->media_flags, &other_media->media_flags, MEDIA_FLAG_RTCP_MUX);
 


rfuchs:
Can you describe a concrete case that this would fix? I would like to add it to the automated tests.

jakubkarolczyk:
Client sends the offer with rtcp-mux; RTPEngine does rtcp-demux and sends the new offer to peer
1st answer back from the peer (183) with c=A.B.C.D; RTPEngine replies to client with rtcp-mux
2nd answer back from the peer (200) with c=E.F.G.H; RTPEngine replies to client without rtcp-mux (before fix)
Client can only do rtcp-mux and rejects the call

2nd answer back from the peer (200) with c=E.F.G.H; RTPEngine replies to client with rtcp-mux (after fix)

rfuchs:
Hm ok, can't really reproduce that... does the 200 answer also have a different to-tag than the 183?

jakubkarolczyk:
That's correct, 200 has got different to-tag than 183

rfuchs:
Ok I see. I don't think this is the correct solution. In fact it breaks an existing test case.

Is there any branching/forking going on in your routing logic, or do you simply get two answers with different to-tags unsolicited?

jakubkarolczyk:
I simply get two answers with different to-tags.

btw. what is the scenario of test case "gh #793 b2"?

rfuchs:
> I simply get two answers with different to-tags.
> 
> btw. what is the scenario of test case "gh #793 b2"​【5 cm】?

Similar to your case, but for a deliberately branched call (using `via-branch` and multiple offers).

I'll have to think about how this can be fixed nicely.

rfuchs:
stale

PR #1127, opened 2020-12-04
Update jitter_buffer.c

jakubkarolczyk:
Changed the way of excluding RTCP from being processed by JB.
For rtcp-mux, stream is both RTP and RTCP. That causes the situation when the first RTCP packet received is disabling Jitter Buffer. The proposition is to rely on check every packet and exclude the RTCP one from being processed in JB.


rfuchs (commented):
Ok generally, but see my comments. Also I would appreciate a more descriptive primary line in the commit (instead of just what file it changes), thanks 👍

PR #1133, opened 2020-12-11
Implement RFC7714 ciphers: AEAD AES

space88man:
Initial userspace only implementation. kernel implementation to follow.


space88man:
Question for @rfuchs : the AEAD-AES GCM encrypt operations increases the size output. The API `crypto_encrypt_rtp/crypto_encrypt_rtcp` look like in-place operations, can the source == target str accept the increased length of str?

rfuchs:
To answer your question, there's a `#define RTP_BUFFER_TAIL_ROOM	512` to accommodate for increased payload length during rewrite operations. You can increase it if necessary. On the kernel side we have `#define MAX_SKB_TAIL_ROOM (sizeof(((struct rtpengine_srtp *) 0)->mki) + 20)`

rfuchs:
And also yes, the tests will have to be adapted to expect the additional crypto suites.

rfuchs:
Currently working on cleaning this up and seeing if kernel support is feasible

PR #1159, opened 2021-01-08
Play and repeat

dvillaume:
Add a repetitor to the play functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1795,6 +1795,10 @@ Media files can be provided through one of these keys:
 	database via (at the minimum) the `mysql-host` and `mysql-query` config keys. The daemon will then
 	retrieve the media file as a binary blob (not a file name!) from the database via the provided query.
 
+* `repeat_times`
+
+    Contains an integer. File will be repeated in consequence (default is 1).
+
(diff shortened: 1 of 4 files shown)


rfuchs (commented):
Overall looks good to me, with a few comments, see below. My primary concern is with the usage of `avformat_seek_file` and then there's a few minor style issues.

Lastly, since this is a single feature/change, all these commits should be squashed into a single one, together with a more descriptive commit messages describing the change.

Since these are all minor issues, I can handle them if you prefer. Let me know.

dvillaume:
@rfuchs i will accept your poposal to handle this, if that doesn't bother you too much.

rfuchs:
So ac2a6f07cb would be my cleaned up version of this. See if it works for you. (Flag has been renamed to `repeat-times` from `repeat_times` for consistency)

amessina:
While not related directly to repeat, it would be useful to play a list of files in succession, i.e. play-media `file=file1&file2&...1` `db-id=id1&id2&...`.

rfuchs:
@amessina pull requests welcome :smile:

amessina:
> @amessina pull requests welcome

Most certainly -- I just don't know C :(

PR #1164, opened 2021-01-15
Update ports binding logic

smititelu:
Right now, when binding ports, same number of ports are bound on all local interfaces of a logical interface. Example of audio only call:

```interface = external/195.1.1.1;external/2001:8d8::1``` will bind 4 ports on IP4 and 4 ports on IP6 even though there is only an IP4 call; thus IP6 ports are wasted.

Try to save ports and bind _only the first available local interface of desired family_, of a logical interface.

Let me know what you think of this approach.

Thanks,
Stefan

### DIFF
--- a/daemon/media_socket.c
+++ b/daemon/media_socket.c
@@ -911,40 +911,84 @@ fail:
 }
 
 /* puts a list of "struct intf_list" into "out", containing socket_t list */
-int get_consecutive_ports(GQueue *out, unsigned int num_ports, const struct logical_intf *log,
-		const str *label)
+int get_consecutive_ports(GQueue *out, unsigned int num_ports, struct call_media *media)
 {
 	GList *l;
 	struct intf_list *il;
 	const struct local_intf *loc;
+	const struct logical_intf *log = media->logical_intf;
+	const sockfamily_t *desired_family = media->desired_family;
+	const str *label = &media->call->callid;
 
+	/*
+	// debug locals of logical incerface
+	char ip[100];
 	for (l = log->list.head; l; l = l->next) {
 		loc = l->data;
+		inet_ntop(loc->spec->local_address.addr.family->af, &loc->spec->local_address.addr.u, ip, sizeof(ip));
+		ilog(LOG_DEBUG, "XXXXXXXXXX IP: %s", ip);
+	}
+	ilog(LOG_DEBUG, "");
+	*/
 
-		il = g_slice_alloc0(sizeof(*il));
-		il->local_intf = loc;
-		g_queue_push_tail(out, il);
-		if (G_LIKELY(!__get_consecutive_ports(&il->list, num_ports, 0, loc->spec, label))) {
-			// success - found available ports on local interfaces, so far
-			continue;
+	if (!rtpe_config.save_interface_ports) {
+		for (l = log->list.head; l; l = l->next) {
+			loc = l->data;
+
+			il = g_slice_alloc0(sizeof(*il));
+			il->local_intf = loc;
+			g_queue_push_tail(out, il);
+			if (G_LIKELY(!__get_consecutive_ports(&il->list, num_ports, 0, loc->spec, label))) {
+				// success - found available ports on local interfaces, so far
+				continue;
+			} else {
+				// fail - did not found available ports on at least one local interface
+				goto error_ports;
+			}
 		}
 
-		// error - found at least one local interface with no ports available
-		goto error_ports;
-	}
-
-	return 0;
+		return 0;
 
 error_ports:
-	ilog(LOG_ERR, "Failed to get %d consecutive ports on all locals of logical '"STR_FORMAT"'",
-		num_ports, STR_FMT(&log->name));
+		ilog(LOG_ERR, "Failed to get %d consecutive ports on all locals of logical '"STR_FORMAT"'",
+			num_ports, STR_FMT(&log->name));
 
-	// free all ports alloc'ed so far for the previous local interfaces
-	while ((il = g_queue_pop_head(out))) {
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: daemon/media_socket.c [hunk 1]
```diff
@@ -911,39 +911,41 @@ int __get_consecutive_ports(GQueue *out, unsigned int num_ports, unsigned int wa
...
 	const struct local_intf *loc;
+	const struct logical_intf *log = media->logical_intf;
+	const sockfamily_t *desired_family = media->desired_family;
+	const str *label = &media->call->callid;
 
 	for (l = log->list.head; l; l = l->next) {
 		loc = l->data;
 
+		// check desired family of local interface
+		if (strcmp(desired_family->rfc_name, loc->spec->local_address.addr.family->rfc_name) != 0)  {
```
rfuchs, line 927: You can directly compare the "family" pointers with == or != since these are static structs. No need to use strcmp.

This is off the top of my head, but the `log` you get here should depend on what `desired_family` was, and the first interface in the `log` list should always be of the correct family. So I _think_ this check can be skipped altogether.

rfuchs (commented):
I haven't tested it, but I suspect that this will break ICE. When ICE is in use, you do want to allocate ports from all interfaces, and then each port will be advertised as an ICE candidate. This is the reason why this is done the way it's done.

Now when ICE is not in use, there really isn't a reason to allocate ports from more than one interface, so only in that case can something like this be done. So this needs to be made conditional on whether ICE is in use. Also switching from a non-ICE endpoint to ICE must be supported, e.g. if only one set of ports is allocated at first but then ICE comes into play later, the remaining ports must be allocated.

So I think a better approach would be to tell `get_consecutive_ports` how many ports/interfaces it should allocate at most, and set it to 1 if ICE is not in use. And then add a check to call it again if ICE is in use and not all ports have been allocated, so it can allocate the rest.

daemon/media_socket.c
smititelu: Thanks for the feedback!

I've been looking into this and I have trouble deciding for port allocation based on ICE presence, for `other_media`, since ICE info comes later, in "answer", whereas ports get already allocated at "offer" time, in `__wildcard_endpoint_map`. I also tried forcing `__get_endpoint_map` for `other_media`, at "answer" time, and (all) ports get allocated, but ICE info is not updated in the SDP of 200OK.

I wonder if this simpler approach is acceptable:
1. Add a new config parameter "save_local_ports", default disabled
2. If disabled, nothing changes in port allocation logic
3. If enabled, alloc ports only on first available local interface, of desired_family
4. Document new config parameter

Thus existing ICE behavior should not change by default.

Thank you,
Stefan

rfuchs: > Thanks for the feedback!
> 
> I've been looking into this and I have trouble deciding for port allocation based on ICE presence, for `other_media`, since ICE info comes later, in "answer", whereas ports get already allocated at "offer" time, in `__wildcard_endpoint_map`. I also tried forcing `__get_endpoint_map` for `other_media`, at "answer" time, and (all) ports get allocated, but ICE info is not updated in the SDP of 200OK.

Hm, is that so? AFAICT the `ICE` flag for both sides of the call should be available and properly set after `__ice_offer` - For the `other_media` it simply depends on what attributes were set in the SDP, and for `media` it depends on the SDP and what option flags were given. But I haven't looked at the whole thing in detail yet so I could be wrong of course.

> 
> I wonder if this simpler approach is acceptable:
> 
> 1. Add a new config parameter "save_local_ports", default disabled
> 2. If disabled, nothing changes in port allocation logic
> 3. If enabled, alloc ports only on first available local interface, of desired_family
> 4. Document new config parameter
> 
> Thus existing ICE behavior should not change by default.

Ok as a workaround - I'd prefer a more versatile and complete solution, but we can use this as a stepping stone, and then maybe I can finish it off when I have the time to play with it.

smititelu:
Updated PR to the latest master, with my commit on top, to check if sid debian packaging still fails.

smititelu:
Check was still failing for sid, for latest master, even though compilation and tests passed.

After trying a "blind" commit over the latest master (which just added a comment in code) and force pushed __the same__ "port binding" commit over the latest master, now all checks succeed.

daemon/media_socket.c
```diff
@@ -911,40 +911,84 @@ int __get_consecutive_ports(GQueue *out, unsigned int num_ports, unsigned int wa
...
 
-	return -1;
+		return -1;
+
+	} else {
+		for (l = log->list.head; l; l = l->next) {
+			loc = l->data;
+
+			// check desired family of local interface
+			if (desired_family != loc->spec->local_address.addr.family)  {
```
rfuchs, line 968: Did you have a chance to see if this test is actually necessary? Because I think it should not ever happen. But we can leave it in here for now.

rfuchs (commented):
Yeah some of the tests seem a bit unstable, still not quite sure why...

LGTM overall. Last request would be to add the new option to the man page (.pod) file, thx

daemon/media_socket.c
smititelu: Yes, I've been looking into it and I consider it is needed because:
1. local interfaces from a logical interface are indeed ordered cosidering the desired family, however, it won't contain only local interfaces of that desired family but both families e.g. (IP6 IP6 IP4 IP4, not just IP6 IP6)
2. If IP6 is desired, and no local interface has available ports, I'd like __not__ to bind IP4 ones, but just give error

smititelu:
Updated .pod file.

daemon/media_socket.c
rfuchs: Fair enough, thanks for explanation

PR #1180, opened 2021-02-01
Control ng tcp

enreached:
This patch adds support for control-ng over TCP; we feel this to be an important feature, it was tested in our production environment for some time and the results are encouraging.

### DIFF
--- a/daemon/control_ng.c
+++ b/daemon/control_ng.c
@@ -1,5 +1,6 @@
 #include "control_ng.h"
 
+#include <errno.h>
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <assert.h>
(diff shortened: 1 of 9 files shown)


daemon/control_ng.c
```diff
@@ -463,9 +544,63 @@ struct control_ng *control_ng_new(struct poller *p, endpoint_t *ep, unsigned cha
...
 
+	rand_hex_str(cookie_buf, sizeof(cookie_buf) / 2);
+	control_ng_send(&cookie, to_send, &s->sock.remote, &s->sock);
 }
 
+void notify_ng_tcp_clients(str *data) {
+	mutex_lock(&tcp_connections_lock);
+	g_hash_table_foreach(tcp_connections_hash, notify_tcp_client, data);
+	mutex_unlock(&tcp_connections_lock);
+}
```
rfuchs, line 603: I don't see these being used anywhere. Some kind of leftovers?

rfuchs (commented):
Untested but looks good overall. Just a note that you can also already use the Websocket protocol to do something very similar. Also maybe using something like Netstring encapsulation would make the whole bencode_valid exercise unnecessary.

daemon/control_ng.c
rfuchs: Ah ok, no that's fine then.

rfuchs:
Oh and if you could document the new option, in the readme and the .pod file, that would be great :laughing:

rfuchs:
As the question came up on the Kamailio mailing list (https://lists.kamailio.org/pipermail/sr-users/2021-March/112029.html), out of curiosity, what are you using to speak TCP to rtpengine?

enreached:
We have a custom piece of software which speaks TCP to rtpengine.

PR #1200, opened 2021-03-02
Implement poller per thread in order to fix packet order.

enreached:
With this patch we implemented a poller per thread which will guarantee the correct order of packets being read and sent. In our production environment we noticed that sometimes DTMF packets (RFC 2833) are being sent from RTPEngine in an incorrect order, even though they were delivered in the correct one. Our conclusion was that this was caused by the fact that different threads are reading (and thus sending) DTMF packets from the same socket. Having a poller per thread, we are certain that only a single thread will ever read from a single socket, thus maintaining the correct packet order.

### DIFF
--- a/daemon/poller.c
+++ b/daemon/poller.c
@@ -598,18 +598,6 @@ now:
 	}
 }
 
-static void sleep_ms(int ms) {
-	struct timespec deadline;
-        long next_tick;
-        clock_gettime(CLOCK_MONOTONIC, &deadline);
-
-        next_tick = (deadline.tv_sec * 1000000000L + deadline.tv_nsec) + ms * 1000000;
-        deadline.tv_sec = next_tick / 1000000000L;
-        deadline.tv_nsec = next_tick % 1000000000L;
-
-        clock_nanosleep(CLOCK_MONOTONIC, TIMER_ABSTIME, &deadline, NULL);
-}
-
 void poller_loop(void *d) {
 	struct poller_map *map = d;
 	poller_map_add(map);
@@ -618,7 +606,7 @@ void poller_loop(void *d) {
 	while (!rtpe_shutdown) {
 		int ret = poller_poll(p, 100);
 		if (ret < 0)
-			sleep_ms(20);
+			usleep(20 * 1000);
 	}
 }
 
--- a/daemon/poller.c
+++ b/daemon/poller.c
@@ -11,7 +11,6 @@
 #include <errno.h>
 #include <sys/epoll.h>
 #include <glib.h>
-#include <sys/syscall.h>
 #include <sys/time.h>
 #include <main.h>
 #include <redis.h>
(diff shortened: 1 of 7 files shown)


rfuchs (commented):
Couple of points here.

One side effect of such an approach is that if any poller thread should block for any reason (for example when writing an update to Redis, or when triggering call recording, or doing ffmpeg stuff), then all sockets that are bound to that poller would be affected and might get blocked. Therefore the entire feature should be made optional.

Also you would still probably see the same effect when kernel forwarding is enabled.

I wonder if the implementation could be made simpler? We know ahead of time how many poller threads we'll have, so you could simply create one poller per thread (kept in an array or a list), then launch one thread for each poller, and then assign created sockets (or entire calls) round-robin to individual pollers. Or is there some benefit to going by thread ID like this?

(There's also `SO_INCOMING_CPU` as a possible alternative, but no idea how well that would work for this, if even at all.)

daemon/poller.c
```diff
@@ -537,7 +600,31 @@ void poller_timer_loop(void *d) {
...
+
 void poller_loop(void *d) {
+	struct poller_map *map = d;
+	poller_map_add(map);
+	struct poller *p = poller_map_get(map);
+
+	while (!rtpe_shutdown) {
+		int ret = poller_poll(p, 100);
+		if (ret < 0)
+			sleep_ms(20);
```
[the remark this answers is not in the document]
enreached: I tried to create some kind of an thread affinity, if thread A creates another media socket, then assign itself to that socket by getting a poller from map belonging to the same thread ID.

The purpose of the sleep is to prevent excess CPU usage in _poller_poll_:
<snip>
        mutex_lock(&p->lock);

        ret = -1;
        if (!p->items || !p->items_size)
                goto out;
<snip>

if p->items_size is 0 or p->items is NULL this will bail out and be called immediately again.

If I make this configurable, is there still a chance for you to accept it?

rfuchs:
Ok I see. Yeah I'm willing to accept it if it's made optional. Some other things I noticed here that I don't see as blockers, but JFYI

You could use `pthread_self` for a cleaner and more portable alternative for `gettid`

For random numbers, we tend to use `ssl_random()` in the code, which is a bit shorter

For small sleeps, we tend to use `usleep()` which would eliminate the need for a hand rolled function

You could call `poller_loop2` from `poller_loop` to eliminate some code duplication

enreached:
Last commits should address all the issues you've risen. What should we do about tests? Do you think we should test the behavior with and without the 'poller per thread' functionality?

rfuchs:
> Last commits should address all the issues you've risen. What should we do about tests? Do you think we should test the behavior with and without the 'poller per thread' functionality?

TBH I'm not sure how that can be tested reliably since you may randomly get the same behaviour even without the feature active...

daemon/poller.c
```diff
@@ -50,9 +50,68 @@ struct poller {
...
+struct poller *poller_map_get(struct poller_map *map) {
+	if (!map)
+		return NULL;
+
+	struct poller *p = NULL;
+	pthread_t tid = pthread_self();
+	mutex_lock(&map->lock);
+	p = g_hash_table_lookup(map->table, (gpointer)tid);
+	if (!p) {
+		gpointer *arr = g_hash_table_get_keys_as_array(map->table, NULL);
```
rfuchs, line 91: `arr` needs to be freed with `g_free`

rfuchs (commented):
One more thing I noticed when going through the changes again...

Please document the new option in `rtpengine.pod` and then I can merge this.

daemon/poller.c
enreached: Good catch, thanks.

rfuchs:
Merged as f650f951

PR #1213, opened 2021-03-19
added rtpengine-ctl set and list options for the delete-delay config parameter

hdikme:
The patch enables the following commands:
- rtpengine-ctl list deletedelay
- rtpengine-ctl set deletedelay X (where X is a number).


rfuchs:
Can you also add it to the online help in `rtpengine-ctl` itself please

hdikme:
Compilation and testing steps have been successfully fulfilled on the local. I don't know why it hangs here on the Unit Tests.

hdikme:
@rfuchs thanks for the quick reply, it's done!

PR #1218, opened 2021-03-24
max-mos: is often invalid when RTCP was not received on both legs 

jchavanton:
Since we are using RTT mos-cq (conversational quality), we should not compute it when we have only the RTT from one
leg.

What is happening is that almost everytime you have significant latency, the min-mos is the first one and is not accurate ... 

We should consider this mos-cq invalid.

Further actions :
- add MOS-LQ (the mos that does not consider RTT)
- modify Kamailio rtpengine module to not select mos = 0 as the lower mos.

```
if (decode_mos_vals_dict(&vals_decoded, ssrc_dict, "lowest MOS")) {
        if (vals_decoded.mos > 0 && vals_decoded.mos < min_vals.mos)
                min_vals = vals_decoded;
}
```

Speaking about this, this seems like a good article for more context.
https://archive.eetasia.com/www.eetasia.com/ARTICLES/2005OCT/B/2005OCT03_RFD_NETD_TA.pdf?SOURCES=DOWNLOAD

### DIFF
--- a/daemon/ssrc.c
+++ b/daemon/ssrc.c
@@ -60,6 +60,9 @@ static void ssrc_entry_put(void *ep) {
 
 // returned as mos * 10 (i.e. 10 - 50 for 1.0 to 5.0)
 static void mos_calc(struct ssrc_stats_block *ssb) {
+	if (!ssb->rtt)
+		return; // can not compute the MOS-CQ unless we have a valid RTT
+
 	// as per https://www.pingman.com/kb/article/how-is-mos-calculated-in-pingplotter-pro-50.html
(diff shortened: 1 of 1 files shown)


jchavanton:
looking at the logs, I can not see why the unit tests failed ...

rfuchs:
> looking at the logs, I can not see why the unit tests failed ...
> 
> Seems like it is failing at the very end, not sure I understand the framework
> 
> ```
> 1..1332
> Died at ../perl/NGCP/Rtpengine/AutoTest.pm line 265.
> END failed--call queue aborted.
> # Looks like your test exited with 22 just after 1332.
> ```

That test is currently broken, don't worry about it

PR #1219, opened 2021-03-25
Add stats for ipv4/ipv6/mixed media calls

smititelu:
Also Send stats for ipv4/ipv6/mixed media calls to graphite

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -2115,6 +2115,19 @@ int monologue_offer_answer(struct call_monologue *other_ml, GQueue *streams,
 
 	ml_media = other_ml_media = NULL;
 
+	// reset offer ipv4/ipv6/mixed media stats
+	if (flags && flags->opmode == OP_OFFER) {
+		statistics_update_ip46_inc_dec(call, CMC_DECREMENT);
+		call->is_ipv4_media_offer = 0;
+		call->is_ipv6_media_offer = 0;
+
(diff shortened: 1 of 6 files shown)


rfuchs:
I believe there's a problem if a call has one side switching from v4 to v6 (or vice versa) during a re-invite - the stats would go out of sync then, would they not? In fact it looks like that any time a call sees two or more answers it would be counted multiple times?

smititelu:
Indeed I have overlooked the re-invites. Will update the PR accordingly.

Question: Do you think there is any valid scenario where calling rtpengine_answer() when call is already answered is helpful? Or this should just be guarded out, for these statistics?

Thanks,
Stefan

rfuchs:
> Indeed I have overlooked the re-invites. Will update the PR accordingly.
> 
> Question: Do you think there is any valid scenario where calling rtpengine_answer() when call is already answered is helpful? Or this should just be guarded out, for these statistics?

Yes, this is very commonly done for a number of reasons, and can include changing the endpoint and address family.

smititelu:
I've updated PR to consider separate offer/answer ip 4/6 stats and to update the stats for each new offer/answer.

I've basically tested it and it looks fine. Still, I want to do more complex tests and will post later.

smititelu:
I've found no issues in my tests, so far, with these new statistics. From my point of view PR can be merged.

PR #1222, opened 2021-03-26
Mos average fix

jchavanton:
I noticed that mos average also needs to be fixed when missing one leg RTT, here is one example with the problem.

```
[core] --- SSRC 40658ef2
[core] ------ Average MOS 3.2, lowest MOS 0.0 (at 0:04), highest MOS 4.3 (at 0:08)
[core] --- SSRC 57f548f4
[core] ------ Average MOS 4.3, lowest MOS 4.3 (at 0:05), highest MOS 4.3 (at 0:05)
````

with this fix 

```
[core] Calculated MOS from RR for 40671f96 is 0.0
```
and then end of call is accurate
```
--- SSRC 40671f96
------ Average MOS 4.3, lowest MOS 4.3 (at 0:08), highest MOS 4.3 (at 0:08)
--- SSRC 622e6eb
------ Average MOS 4.3, lowest MOS 4.3 (at 0:05), highest MOS 4.3 (at 0:05)
```

```

### DIFF
--- a/daemon/call_interfaces.c
+++ b/daemon/call_interfaces.c
@@ -1730,7 +1730,10 @@ static void ng_stats_ssrc(bencode_item_t *dict, struct ssrc_hash *ht) {
 		if (!se->stats_blocks.length || !se->lowest_mos || !se->highest_mos)
 			continue;
 
-		ng_stats_ssrc_mos_entry_dict_avg(ent, "average MOS", &se->average_mos, se->stats_blocks.length);
+		int mos_samples = se->stats_blocks.length - se->no_mos_count;
+		if (mos_samples < 1) mos_samples = 1;
+
+		ng_stats_ssrc_mos_entry_dict_avg(ent, "average MOS", &se->average_mos, mos_samples);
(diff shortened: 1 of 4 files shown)


daemon/call_interfaces.c
```diff
@@ -1730,7 +1730,7 @@ static void ng_stats_ssrc(bencode_item_t *dict, struct ssrc_hash *ht) {
 		if (!se->stats_blocks.length || !se->lowest_mos || !se->highest_mos)
 			continue;
 
-		ng_stats_ssrc_mos_entry_dict_avg(ent, "average MOS", &se->average_mos, se->stats_blocks.length);
+		ng_stats_ssrc_mos_entry_dict_avg(ent, "average MOS", &se->average_mos, (se->stats_blocks.length - se->no_mos_count));
```
rfuchs, line 1733: This may end up dividing by zero now - please add safeguards to protect from that.

jchavanton: very true

jchavanton: just pushed the safety and made normal condition tests

PR #1225, opened 2021-04-02
fix rtt from rtcp_rx

jchavanton:
I found out we where using the RTT from RTCP_XR (both legs in ms)
to represent 1 leg RTT us.


daemon/ssrc.c
```diff
@@ -356,10 +356,14 @@ void ssrc_receiver_report(struct call_media *m, const struct ssrc_receiver_repor
 
 	ilog(LOG_DEBUG, "Adding opposide side RTT of %u us", other_e->last_rtt);
 
+	long long rtt_end2end = other_e->last_rtt ? (rtt + other_e->last_rtt) : 0;
+	if (e->last_rtt_xr > 0) { // use the RTT from RTCP-XR (in ms)
```
rfuchs, line 360: `e` is undeclared here

PR #1262, opened 2021-05-10
Support Configurable MTU

### LINKED ISSUE #1261: Support configurable MTU
We encountered a problem with RTPEngine setting the MTU to 1500 but a client network had set their MTU/MSS to 1380. We saw DTLS packets being dropped in their network causing voice to be unable to establish. Ideally we would want to support Path MTU discovery, but we can't trust that ICMP won't be blocked externally to our network. Therefore, we think the best option is to make the MTU configurable when RTPEngine is compiled to set the MTU. Defaulting to 1200 as that is what Chrome uses - https://groups.google.com/g/discuss-webrtc/c/gH5ysR3SoZI/discussion.

tinoIII:
- default to 1200
- minmimum value 576
- RESOLVE #1261

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -478,7 +478,7 @@ static void options(int *argc, char ***argv) {
 		{ "jb-clock-drift",0,0,	G_OPTION_ARG_NONE,	&rtpe_config.jb_clock_drift,"Compensate for source clock drift",NULL },
 		{ "debug-srtp",0,0,	G_OPTION_ARG_NONE,	&debug_srtp,		"Log raw encryption details for SRTP",	NULL },
 		{ "dtls-rsa-key-size",0, 0,	G_OPTION_ARG_INT,&rtpe_config.dtls_rsa_key_size,"Size of RSA key for DTLS",	"INT"		},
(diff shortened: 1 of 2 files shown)


daemon/main.c
```diff
@@ -477,6 +478,7 @@ static void options(int *argc, char ***argv) {
 		{ "jb-clock-drift",0,0,	G_OPTION_ARG_NONE,	&rtpe_config.jb_clock_drift,"Compensate for source clock drift",NULL },
 		{ "debug-srtp",0,0,	G_OPTION_ARG_NONE,	&debug_srtp,		"Log raw encryption details for SRTP",	NULL },
 		{ "dtls-rsa-key-size",0, 0,	G_OPTION_ARG_INT,&rtpe_config.dtls_rsa_key_size,"Size of RSA key for DTLS",	"INT"		},
+		{ "dtls-mtu",0, 0,	G_OPTION_ARG_INT,&rtpe_config.dtls_mtu,"DTLS MTU)",	"INT"		},
```
rfuchs, line 481: Looks good except for this little typo here (extra `)`) :smile: 

Also please document the new option in the .pod file

Issue #1258, opened 2021-05-07
Foreign sessions are stopped being established after some time of running


pkuzak:
Foreign sessions are stopped being established after some time of running. It is difficult for me to give more details and it difficult for me to reproduce this. But running it in our live system, after some minutes there is no change on foreign sessions anymore. The established ones are slowly terminated by timeout, but no new ones are established. 
I saw in the commits that some work was done in this context, maybe you have a clue what change could cause this behavior.

rfuchs:
Not off the top of my head. As long as the Redis key notifications are working, you should see new foreign sessions being established. You can verify that by looking at the data in Redis. Also sessions marked as "foreign" should be exempt from any timeouts (save the "final" timeout).

PR #1268, opened 2021-05-20
Avoid deadlock of notification thread

smititelu:
Fixes issue #1258.

Testing the other two commits on this branch (rfuchs/1257) also look good for us, so they can be put on master, with this commit included.


rfuchs:
Thanks for the help. Merged as 156271ef

PR #1285, opened 2021-06-17
Fix G729 packet duration

semhoun:
Without this correction, timestamp is not incremented in case of transcoding


rfuchs:
Merged as 5b90489fc30a1337b9b43df82b14e5303792daa4

PR #1307, opened 2021-07-13
Removing unused function '__unsubscribe_one'

HauptmannBoosted:
In daemon/call.c the function '__unsubscribe_one'
has been removed because it was never used which
produces gcc warning on compilation:

> call.c:2611:13: warning: ‘__unsubscribe_one’ defined but not used [-Wunused-function]
>  2611 | static bool __unsubscribe_one(struct call_monologue *which, struct call_monologue *from) {
>      |             ^~~~~~~~~~~~~~~~~


rfuchs:
I have commits pending which will use this function in the near future. Sorry for the warnings in the meantime and sorry for the trouble.

Issue #1310, opened 2021-07-16
Dropped packet in DTLS handshake causes DTLS to fail


jamesaylett:
**Defect summary**

When acting as the DTLS server, if the final DTLS packet sent from rtpengine is lost, then rtpengine does not respond to retransmitted client certificate/key exchange messages.

**Observed behavior**

Client sends hello, rtpengine responds with server hello, client sends certificate/key exchange, rtpengine new session ticket, client then repeatedly sends certificate/key exchange.

The rtpengine debug logs show "DTLS handshake successful" followed by repeated "Processing incoming DTLS packet".

The call has no audio.

**Expected behavior**

When client sends retransmissions, rtpengine responds by sending DTLS and is able to successfully complete DTLS handshake. Audio is heard on both legs.

**To reproduce**

Use [clumsy](http://jagt.github.io/clumsy/download.html) tool (or similar) to simulate an unreliable network, dropping the final DTLS packet from rtpengine on the client using source/destination IP filters and packet length to drop the single packet from rtpengine in the DTLS exchange as described above. Then removing the packet filtering when the client sends retransmissions.

**Logs**

I can upload logs/pcap traces if needed.

**Environment**

Centos 7.6, rtpengine master, OpenSSL 1.1.1k
Centos 7.6, rtpengine 9.3.1, OpenSSL 1.0.2k

**Resolution**

See pull request, I suggest modifying try_connect() to call SSL_read() when connected flag is set instead of returning early.

PR #1311, opened 2021-07-16
Handle dtls retransmissions

jamesaylett:
See issue https://github.com/sipwise/rtpengine/issues/1310 for summary.

I'm not too familiar with OpenSSL, but taking examples from other sources such as Mediasoup lead me to believe calling SSL_read()  anytime after SSL_connect/SSL_accept would resolve and handle retransmissions.

### DIFF
--- a/daemon/dtls.c
+++ b/daemon/dtls.c
@@ -479,13 +479,13 @@ int dtls_verify_cert(struct packet_stream *ps) {
 
 static int try_connect(struct dtls_connection *d) {
 	int ret, code;
-
-	if (d->connected)
-		return 0;
+	unsigned char buf[0x10000];
 
 	__DBG("try_connect(%i)", d->active);
 
-	if (d->active)
+	if (d->connected)
+		ret = SSL_read(d->ssl, buf, sizeof(buf)); /* retransmission after connected - handshake lost */
+	else if (d->active)
 		ret = SSL_connect(d->ssl);
 	else
 		ret = SSL_accept(d->ssl);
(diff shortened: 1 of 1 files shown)


rfuchs:
Sounds reasonable. Have you tested/verified that this solution actually fixes the problem?

jamesaylett:
Apologies for not answering sooner.

Our QA team have been looking at the code and have found some in some circumstances that these changes don't have the desired effect, likely when packets are coming in out of order. We believe try_connect may be exiting with SSL_WANT_READ/SSL_WANT_WRITE and subsequently causing an SSL invalid operation when SSL_read is called. I am going to follow up, so in the meantime please don't merge - but may be helpful to keep this open for reference to others.

My own testing did not have any issues, but was not done under load and with the latest OpenSSL version 1.1.1k.

In terms of the severity of this issue, its not clear to us. We normally get a few customer reports of no-way speech from time to time. Over internet, packets are more likely to be dropped. Could be 1 in 10,000 or less.

rfuchs:
Thanks for trying to track down the issues.

I was wondering if simply calling SSL_connect or SSL_accept again would take care of it? Using SSL_write seems a bit odd to me as we're not reading/writing any actual SSL data. But then I'm not too familiar with the OpenSSL internals.

jamesaylett:
I did try repeating the call to SSL_connect, but in my tests it wasn't successful. I think this does make sense - the DTLS negotiation is incomplete and the documentation on [SSL_connect](https://www.openssl.org/docs/man1.1.1/man3/SSL_connect.html) implies its use is to initiate a handshake.

One option might be to change the underlying BIO to block, the OpenSSL documentation suggesting that SSL_connect/SSL_accept returns once the handshake is complete. Not being familiar with OpenSSL, I wasn't sure whether it is an option for DTLS and whether blocking IO would have negative impact on RTPEngine's ability to handle call volumes.

For more context, we use RTPEngine exclusively for WebRTC to legacy VoIP handling multiple thousands concurrent calls daily. Every so often we get these reports of no way speech, but tracking down the exact cause is often too time consuming to be worthwhile given the low rate of occurrence. We now have a customer who is more sensitive to failed calls, so this is an area we are looking to improve to help with the robustness of our service.

I'll likely take a few days to investigate, but I will be sharing the results. Thank you.

jamesaylett:
Apologies for leaving this for so long, been pulled onto other priorities but now will be looking at picking this up again. 

We repeated our QA and found for OpenSSL 1.0.2k this pull request resolved the issue, however with 1.1.1k the issue seemed to be resolved with some builds but on others we were able to reproduce the issue about 50% of the time. With our 1.1.1k build we are seeing some compiler warnings (using CentOS 7 so the OS OpenSSL is 1.0.2), I will investigate further and get back to you!

PR #1373, opened 2021-10-08
Do not manage TLS sockets unless we are forwarding to them

vaizki:


vaizki:
> I'm confused about how `(stream->forwarding_on || mf->forwarding_on)` can become true if TLS forwarding isn't configured?

`stream->forwarding_on` becomes true for half the streams in a call after a `start forwarding` message is sent to rtpengine, metafile gets FORWARDING section at that point.

https://github.com/sipwise/rtpengine/blob/master/recording-daemon/metafile.c#L193

vaizki:
> Ok that's what I thought. AFAIK `start forwarding` is about TLS forwarding, not the Unix socket forwarding.

Oh. Maybe I shouldn't do this at 3 am... So unix socket forwarding is always on for all recorded calls and applies to all streams.  Looks like I need find another way to forward only part of the streams. If I close the socket from the other side, there is a constant stream of error-level messages to the log for the rest of the call (one message per RTP packet). So I need to consume every packet even when I know from metadata that I don't need this call.

This PR is dead then, instead there should be an error message if forwarding is started without a valid TLS config.

rfuchs:
You can open a PR to add support for start/stop forwarding for Unix socket forwarding

PR #1400, opened 2021-11-17
Add support for endpoint-learning on a per call basis

john08burke:
Currently, the `endpoint-learning` algorithm is a global config and set at startup.  This PR adds the ability to override the startup config on a per-call basis.  The reason behind wanting per-call control of the EL algo was driven by [this discussion](https://groups.google.com/g/rtpengine/c/AUJRiMnEcJc) on the mailing list.  In short, I found that on some of the newer kernels, userspace can be quite CPU expensive when compared to older versions.  By having some per-call control, this gives the developer the ability to optimize CPU allocation by minimizing / eliminating any userspace processing (ex: traffic that is known to advertise public SDP doesn't need 3 sec user learning phase).

This is my first PR on this project, so let me know if I was way off base with anything... or if you want anything refactored!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1184,6 +1184,10 @@ Optionally included keys are:
 		When processing a non-OSRTP answer in response to an OSRTP offer, accept the
 		OSRTP offer anyway. Results in RTP/SRTP transcoding.
 
+* `endpoint-learning`
+
+	Contains one of the strings `off`, `immediate`, `delayed` or `heuristic`. This tells rtpengine which endpoint learning algorithm to use and overrides the `endpoint-learning` configuration option. This option can also be put into the `flags` list using a prefix of `endpoint-learning-`.
+
 * `record call`
 
 	Contains one of the strings `yes`, `no`, `on` or `off`. This tells the rtpengine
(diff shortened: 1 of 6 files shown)


rfuchs:
Hi,

First off, thanks for adding this feature.

Looking through the code, I think this commit is a lot more complicated than it needs to be :)

My suggestions would be:

1) Leave `enum endpoint_learning` as it is (in `main.h`), including the option `struct rtpengine_config` (except maybe for renaming it). This would then be the default value, and take the role of `el_option_def`. The same enum type can then also be used in `struct sdp_ng_flags` without having to redefine it.
2) Use that same enum type for `el_flags` in `struct packet_stream`. There's no need for this to be a bit field as only one option can be set at any given time. This eliminates the `PS_EL` defines and the bit field test stuff.
3) Initialising `el_option` in `call_ng_flags_init()` from the default setting (would then be `rtpe_config.endpoint_learning`) is fine.
4) If you want to be extra fancy (optional exercise for the reader 😆), you can unify the string parsing done in `options()` with the parsing done in `ng_el_option()` into a single function to eliminate code duplication.

john08burke:
Hey @rfuchs, just pushed updates w/ your recommendations.  Indeed a much simpler solution!  My tests checked out, so let me know if any additions mods are required!

PR #1403, opened 2021-11-22
Various changes of RTP timeout handling to handle timeouts on single streams

arnd-s:
- introduced timeout-mode parameter to control mode of operation off/any/all during session
  off - don't monitor timeouts / turn monitoring off
  any - react if a timeout occur on any RTP stream
  all - react if timeout occur on all streams (old behavior)
- Don't monitor RTCP streams for timeout (to prevent false alarms if RTP timeout is very low)

This patch introduces primary the possibility to react on RTP timeouts of a single stream.

During call establishment / early-media, there is often only one stream direction active. To prevent false alarms
for this case, the timeout-mode parameter for the call-interface was added. The logic is, that a call starts with
a disabled timeout monitoring and when the session is established in both directions (for example 200OK Reply for
an answer) the timeout will be activated.

When activated, timeout_activated element of struct call will be set to earliest possible timeout occurence
(activation time + timeout time). This should prevent false alarms immediatliy after session establishment

There is still an unresolved problem at the moment. Even timeout_activated is used, it sometimes happens
that on first check a timeout is detected. This occurs even, when timeout value (and timeout_activated
in succession) is increased.
To circumvent this, missed_packet_counter of struct packet_stream is used, so that the checks fails on the third
missed package in a row.

Signed-off-by: Arnd Schmitter <arnd@lgcm.de>

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -145,7 +145,8 @@ void call_make_own_foreign(struct call *c, bool foreign) {
 static void call_timer_iterator(struct call *c, struct iterator_helper *hlp) {
 	GList *it;
 	unsigned int check;
-	bool good = false;
+	bool good = false; // gets true if any stream has active media
+	bool failed = false; // gets true if any stream has a timeout event
 	struct packet_stream *ps;
 	struct stream_fd *sfd;
 	int tmp_t_reason = UNKNOWN;
@@ -175,6 +176,12 @@ static void call_timer_iterator(struct call *c, struct iterator_helper *hlp) {
 	if (IS_FOREIGN_CALL(c)) {
 		goto out;
 	}
+
+	if (c->timeout_mode == TIMEOUT_OFF)
+		goto out;
+
+	if (c->timeout_activated > rtpe_now.tv_sec)
+		goto out;
 
 	if (c->deleted && rtpe_now.tv_sec >= c->deleted
 			&& c->last_signal <= c->deleted)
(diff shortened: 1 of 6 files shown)


daemon/call.c
```diff
@@ -206,6 +213,16 @@ static void call_timer_iterator(struct call *c, struct iterator_helper *hlp) {
 
 		/* valid stream */
 
+		// ignore RTCP Streams
+		if (PS_ISSET(ps, RTCP))
```
rfuchs, line 217: I think it might be more appropriate to use `if (!PS_ISSET(ps, RTP))` here? (Streams can be both RTP and RTCP in case of RTCP-mux)

daemon/call.c
```diff
@@ -4022,6 +4047,7 @@ int call_delete_branch(const str *callid, const str *branch,
 				"(via-branch '" STR_FORMAT_M "') in %d seconds",
 				STR_FMT_M(&ml->tag), STR_FMT0_M(branch), delete_delay);
 		ml->deleted = rtpe_now.tv_sec + delete_delay;
+		ml->mark_deleted = 1;
```
rfuchs, line 4050: I'm not 100% sure, but could `ml->deleted != 0` be used as a test for this instead of adding a new flag?

daemon/call.c
arnd-s: Yes, this sounds better. I hadn't RTCP-mux in mind, because it's not used in our case.

daemon/call.c
arnd-s: As far as i remember i first tried it this way but there was an issue with it but i currently don't know what it was.
I need to make some tests and see, if i can  reproduce the issues i had.

rfuchs:
stale

PR #1407, opened 2021-12-02
RTPengine remote syslog server 

daknadnane:
Hi , 

is  possible to configure rtpengine to send log to remote syslog server


rfuchs:
No, not directly. You can do this through your system's syslog config.

Questions should be directed to our mailing list: https://rtpengine.com/mailing-list

PR #1413, opened 2021-12-16
Add 'seconds' to some stats duration

smititelu:


rfuchs:
You need to adjust one of the tests to expect the new format, e.g. at https://github.com/sipwise/rtpengine/blob/master/t/test-stats.c#L431 and https://github.com/sipwise/rtpengine/blob/master/t/test-stats.c#L441 and all other instances in that file.

PR #1428, opened 2022-01-11
Add new graphite-timeout parameter

smititelu:
Needed to be able to set graphite socket timeout.
Useful when one wants rtpengine to force the graphite connection
to fail faster, in case graphite server gets filtered while
connection is ongoing.

### DIFF
--- a/daemon/graphite.c
+++ b/daemon/graphite.c
@@ -56,6 +56,11 @@ static int connect_to_graphite_server(const endpoint_t *graphite_ep) {
 	ilog(LOG_INFO, "Connecting to graphite server %s", endpoint_print_buf(graphite_ep));
 
 	rc = connect_socket_nb(&graphite_sock, SOCK_STREAM, graphite_ep);
+
+	if (rtpe_config.graphite_timeout > 0 && !(graphite_sock.fd < 0)) {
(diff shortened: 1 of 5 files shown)


lib/socket.c
```diff
@@ -742,6 +742,20 @@ int connect_socket_nb(socket_t *r, int type, const endpoint_t *ep) {
 	return connect_socket_retry(r);
 }
 
+int connect_socket_nb_timeout(socket_t *r, int type, const endpoint_t *ep, unsigned int timeout) {
```
rfuchs, line 745: Suggestion: connect_socket_nb() could now be turned into a wrapper around connect_socket_nb_timeout() passing a zero timeout to eliminate some code duplication.

Alternatively usertimeout() could simply be called after the socket has been created, explicitly right after connect_socket_nb(). AFAICT it has no effect on connection progress and so setting the option after connect() should be fine, also eliminating the code duplication.

rfuchs (commented):
Please document the new option in the .pod file, thanks.

smititelu:
Thanks for the feedback! PR updated.

I've tested setting socket option right after connect_socket_nb() and it works, even when graphite server is initially filtered.

PR #1429, opened 2022-01-11
Enabled sending total statistics to graphite

hdikme:
Enabled sending also the total statistics to graphite by a config variable called graphite_send_totalstats.

### DIFF
--- a/daemon/graphite.c
+++ b/daemon/graphite.c
@@ -193,6 +193,10 @@ GString *print_graphite_data(void) {
 			(double) atomic64_get(&rtpe_stats_gauge_graphite_min_max_interval.avg.ng_command_times[i]) / 1000000.0);
 	}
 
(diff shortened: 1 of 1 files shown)


rfuchs:
This is only about the NG command stats, isn't it? And they're not split up by proxy, so this metric should already be available without having to go through the list. For a total count of all commands you should look at `rtpe_stats_cumulative.ng_commands[command_idx]`. So something like this (untested)

```
	if (rtpe_config.graphite_send_totalstats) {
		for (int i = 0; i < NGC_COUNT; i++) {
			GPF("total%scount %" PRIu64, ng_command_strings_short[i], atomic64_get(&rtpe_stats_cumulative.ng_commands[i]`));
		}
	}
```

Also I think it should be safe to always add these to the output without needing a new config option 🤷

PR #1430, opened 2022-01-12
Fix `subscribe request` SDP media direction

razvancrainea:
When building the SDP for a `subscribe request` command, take into
consideration the media direction of the source stream - if stream is
`recvonly`, then we do not have anything to send, thus the direction
should be advertised as `inactive`, rather than `sendonly`.

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -2972,7 +2972,7 @@ static int monologue_subscribe_request1(struct call_monologue *src_ml, struct ca
 
 		codec_handlers_update(dst_media, src_media, flags, sp);
 
-		if (MEDIA_ISSET(src_media, SEND))
+		if (MEDIA_ISSET(src_media, RECV))
 			MEDIA_SET(dst_media, SEND);
 		else
 			MEDIA_CLEAR(dst_media, SEND);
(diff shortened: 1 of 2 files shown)


daemon/call.c
```diff
@@ -2972,7 +2972,10 @@ static int monologue_subscribe_request1(struct call_monologue *src_ml, struct ca
 
 		codec_handlers_update(dst_media, src_media, flags, sp);
 
-		MEDIA_SET(dst_media, SEND);
+		if (MEDIA_ISSET(src_media, SEND))
```
rfuchs, line 2975: The SEND/RECV flags are from rtpengine's POV, so I think the destination SEND should be set when the source RECV is set.

razvancrainea: Oh, sorry, I honestly set it empirically, and it was working fine for my use case. Was not aware of the project's workflow, but I am learning it. Will push a fix soon.

rfuchs:
No worries, thanks. Do you want to add some tests for this as well? If not then I'll add them myself.

razvancrainea:
Sure, I'll give it a try.

razvancrainea:
Just committed a test for pausing and resume a single stream. Let me know if that's OK.

rfuchs:
Perfect, thank you

PR #1440, opened 2022-02-08
Run services as non root user

avoylenko:
*) Run services as a user created during **RPM** installation

*) Add **perl-Config-Tiny** dependency

*) Disable automatic kernel module loading for RHEL 7


rfuchs:
Thanks for taking care of the RPM packaging.

FYI we also include non-NGCP packaging for Debian now, so you could omit the `ngcp` prefixes from the RPM packages if you prefer.

PR #1445, opened 2022-02-11
Add more info about ok/failed t tests

smititelu:


rfuchs:
You sure about this logic? This seems to make the result conditional on the exit code from the `echo` and not the test itself.

smititelu:
Thanks for the notice! Indeed there is a flaw in my logic. I am testing now by forcing a test to fail inside the code (and not just inside t/Makefile).

I want to ask you if there is really a need to do "exit 1" when a test fails and stop building final deb?

Thank you,
Stefan

smititelu:
No, all tests run ok.

What I am also trying to achieve, besides adding fail/succeed info for each test, is to run all tests, no matter how many fail (i.e. don't "exit 1" at first failed test). Thus, output can be parsed and a summary can be done.

Let me know what you think of the PR right now. All tests should run. Build should fail if at least one test fails. Tested it by forcing 1 tests to fail in code (aes-crypt.c).

Thanks,
Stefan

rfuchs:
Ok that looks good. If you want to be even more verbose, you could collect the names of the failed tests in a variable and then report them all at the end in one go. Something like
```
failed=""; \
...
else echo `date +"%Y-%m-%d %H:%M:%S"` test $$x: failed ; failed="$$failed $$x"; \
...
if [ -n "$$failed" ]; then \
echo "Tests failed: $$failed"; \
exit 1; \
fi
...
```
should work I think

smititelu:
Updated PR. Tested it with 2 failing tests and it works.

PR #1449, opened 2022-02-14
More info about kernelized stream

smititelu:
Adds call-id and port log prefix, for each kernelized stream. Please let me know what you think.

Thanks,
Stefan

### DIFF
--- a/daemon/media_socket.c
+++ b/daemon/media_socket.c
@@ -1173,6 +1173,8 @@ static const char *kernelize_one(struct rtpengine_target_info *reti, GQueue *out
 	if (sink && !sink->endpoint.address.family)
 		return NULL;
 
+	log_info_stream_fd(stream->selected_sfd);
+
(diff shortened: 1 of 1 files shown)


rfuchs:
Is this because the log prefixes were missing in these log lines? Because that should have been fixed by 6d1dfab77a9ead8aa7be018c60c453e32a37bae1

Or do you want to add extra info? In that case the log_info_stream_fd() should be paired with a log_info_pop() (also see the commit referenced above, https://github.com/sipwise/rtpengine/commit/6d1dfab77a9ead8aa7be018c60c453e32a37bae1#diff-5e71163bdc19a5894ee6f788613376d521032f953485d04d3f35f3d8177be259R36 in particular)

smititelu:
Only `because the log prefixes were missing in these log lines`. Indeed 6d1dfab fixed it. Closing this PR.

Thank you,
Stefan

PR #1457, opened 2022-02-24
provide sendrecv mode in siprec tag-medias and media-labels

razvancrainea:
SIPREC protocol requires to know whether a media stream is actively sending media or not.


rfuchs:
Do you need this merged to any existing branches?

razvancrainea:
I believe it is OK just in sipwise/rtpengine:master for now. Thank you!

Issue #1485, opened 2022-05-02
Crash in DTX processing


szcom:
At 10.4 brach, commit 0f438319d7b0b609a2155ef26fb5a085f212fac8

```
rogram terminated with signal SIGSEGV, Segmentation fault.
#0  0x00005635b1b47721 in __dtx_send_later (ct=0x7f9b745aa580) at codec.c:2877
2877	codec.c: No such file or directory.
[Current thread is 1 (Thread 0x7f9b8aff7700 (LWP 31))]
(gdb) bt
#0  0x00005635b1b47721 in __dtx_send_later (ct=0x7f9b745aa580) at codec.c:2877
#1  0x00005635b1b5623d in timerthread_run (p=0x5635b1bb1020 <codec_timers_thread>) at timerthread.c:62
#2  0x00005635b1ad460d in thread_detach_func (d=0x5635b20c4d00) at aux.c:250
#3  0x00007f9b9e25bfa3 in start_thread (arg=<optimized out>) at pthread_create.c:486
#4  0x00007f9b9e18c4cf in __libc_ifunc_impl_list (name=<optimized out>, array=0x7f9b8aff7700, max=<optimized out>) at ../sysdeps/x86_64/multiarch/ifunc-impl-list.c:385
#5  0x0000000000000000 in ?? ()
(gdb) p ts_dif
No symbol "ts_dif" in current context.
(gdb) p ts_diff
Cannot access memory at address 0x8
(gdb) p *dtxp
$1 = {packet = 0x0, mp = {raw = {s = 0x0, len = 0}, fsin = {address = {family = 0x5635b1baf0c0 <__socket_families>, u = {ipv4 = {s_addr = 1461813054}, ipv6 = {__in6_u = {__u6_addr8 = ">\177!W", '\000' <repeats 11 times>,
              __u6_addr16 = {32574, 22305, 0, 0, 0, 0, 0, 0}, __u6_addr32 = {1461813054, 0, 0, 0}}}}}, port = 1474}, tv = {tv_sec = 1651517653, tv_usec = 61788}, sfd = 0x7f9b68099b80, call = 0x7f9b60f76c40,
[...]

PR #1487, opened 2022-05-04
[dtx] discard null packets in dtx buffer ref #1485

szcom:


daemon/codec.c
```diff
@@ -2911,6 +2911,10 @@ static void __dtx_send_later(struct codec_timer *ct) {
 	while (true) {
 		// do we have a packet?
 		dtxp = g_queue_peek_head(&dtxb->packets);
+	  if (dtxp && !dtxp->packet) {
+		  g_queue_pop_head(&dtxb->packets);
+      continue;
+    }
```
rfuchs, line 2917: I'm not sure this is right. AFAICT the entry with the NULL packet still needs to be processed.

szcom: sure, this PR is just for info to show potential null ptr deref due to null packets in dtx buffer.

since fix for #1485 is sort of incomplete

szcom: @rfuchs in particular 2nd  commit in this pr (eb8ee2763732f73e825f2101980b2509611152c3) shows that __dtx_drift_shift and __dtx_drift_drop can step on nullptr in similar way

rfuchs: Yeah I get that, but I'm saying that simply discarding the entry upon null ->packet is the wrong approach.

PR #1501, opened 2022-06-13
make the build reproducible

linuxmaniac:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1012766

> Whilst working on the Reproducible Builds effort [0] we noticed that
> rtpengine could not be built reproducibly.
>
> Whilst it did use the value from debian/changelog, the RELEASE_DATE
> variable changed due to underlying system timezone setting. A patch
> is attached that passes -u/--utc to the call to date(1).
>
> [0] https://reproducible-builds.org/


mika:
LGTM :+1:

PR #1502, opened 2022-06-14
Add 'all' recording method

smititelu:
Enable both rec methods at the same time. Still under testing.


smititelu:
Found and solved an issue where method 'proc' did not stopped correctly because 'pcap' was renaming metadata file. Decided to just copy it, not rename, when method is 'all'.

Still under testing... pcaps look quite small for the amount of RTP tested.

smititelu:
Updated PR to fix the pcap issue.

The issue was that only a few packets were captured in .pcap, in the beginning, and then pcap stopped growing in size. Fixed it by setting NO_KERNEL_SUPPORT flag for method "all" in "media_socket.c" file.

Testing so far looks good.

include/recording.h
```diff
@@ -85,11 +85,21 @@ struct recording_method {
 };
 
 extern const struct recording_method *selected_recording_method;
+extern const struct recording_method methods[2];
```
rfuchs, line 88: May I suggest a different approach here... instead of using strcmp() all over the place to make the case distinctions, which will inevitably lead to messier and messier code over time, either
1) retain `selected_recording_method` to point to a single recording method, and create a new 3rd recording method "all" (or "both") in the array `methods` that uses wrapper functions to call the functions for both `pcap` and `proc` methods, or
2) eliminate `selected_recording_method` and replace it with something like a bit field, e.g. 0x1 for proc, 0x2 for pcap, 0x3 for both, and then use bit tests in `_rm` to determine which functions to call. The `kernel_support` flag would then have to be carried separately.

daemon/recording.c
```diff
@@ -529,13 +546,20 @@ static int rec_pcap_meta_finish_file(struct call *call) {
...
 	snprintf(new_metapath + prefix_len+fn_len, ext_len+1, ".txt");
-	return_code = return_code || rename(recording->meta_filepath, new_metapath);
-	if (return_code != 0) {
-		ilog(LOG_ERROR, "Could not move metadata file \"%s\" to \"%s/metadata/\"",
-				 recording->meta_filepath, spooldir);
+	if (!strcmp(rtpe_config.rec_method, "all")) {
+		char command[256];
+
+		sprintf(command, "cp %s %s", recording->meta_filepath, new_metapath);
+		return_code = system(command);
```
rfuchs, line 553: Oof... please don't use system() to execute external shell commends. I'll have to look at this in more detail, but isn't the format of the metadata file completely different between the two methods? Are they both pointing to the same file name? If so then perhaps it would be better to make a proper distinction between the two, either in a different path, or using a different file name suffix or something... or simply forgo the `pcap` format file in favour of the `proc` format file in case of `all`.

rfuchs (commented):
Also document this in the man/pod file please.

rfuchs:
LGTM now, let me know how your testing goes.

smititelu:
Over the weekend testing looks good.

PR #1546, opened 2022-10-03
Add TLS ALPN option to be set when connecting MQTT

agranig:
This is required to be set to "mqtt" for instance when connecting to the AWS IoT Core data endpoint at port 443 to indicate we're sending MQTT, because in that case websocket and mqtt shares the same port for whatever reason.

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -559,6 +559,7 @@ static void options(int *argc, char ***argv) {
 #ifdef HAVE_MQTT
 		{ "mqtt-host",0,0,	G_OPTION_ARG_STRING,	&rtpe_config.mqtt_host,	"Mosquitto broker host or address",	"HOST|IP"},
 		{ "mqtt-port",0,0,	G_OPTION_ARG_INT,	&rtpe_config.mqtt_port,	"Mosquitto broker port number",		"INT"},
+		{ "mqtt-tls-alpn",0,0,	G_OPTION_ARG_STRING,	&rtpe_config.mqtt_tls_alpn,	"Mosquitto broker TLS ALPN",	"STRING"},
 		{ "mqtt-id",0,0,	G_OPTION_ARG_STRING,	&rtpe_config.mqtt_id,	"Mosquitto client ID",			"STRING"},
 		{ "mqtt-keepalive",0,0,	G_OPTION_ARG_INT,	&rtpe_config.mqtt_keepalive,"Seconds between mosquitto keepalives","INT"},
 		{ "mqtt-user",0,0,	G_OPTION_ARG_STRING,	&rtpe_config.mqtt_user,	"Username for mosquitto auth",		"USERNAME"},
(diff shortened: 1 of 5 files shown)


rfuchs:
Please also document the option in the .pod file 😆

rfuchs:
Merged as 3a54a1f0a46d151c8b37cfa72484d9ee884edace

linuxmaniac:
@agranig buster build now is broken

From https://kamailio.sipwise.com/view/rtpengine/job/rtpengine-nightly-binaries/architecture=amd64,distribution=buster/116/consoleText
```
mqtt.c: In function 'mqtt_connect':
mqtt.c:65:13: warning: implicit declaration of function 'mosquitto_string_option'; did you mean 'mosquitto_strerror'? [-Wimplicit-function-declaration]
   int ret = mosquitto_string_option(mosq, MOSQ_OPT_TLS_ALPN, rtpe_config.mqtt_tls_alpn);
             ^~~~~~~~~~~~~~~~~~~~~~~
             mosquitto_strerror
mqtt.c:65:43: error: 'MOSQ_OPT_TLS_ALPN' undeclared (first use in this function); did you mean 'MOSQ_OPT_SSL_CTX'?
   int ret = mosquitto_string_option(mosq, MOSQ_OPT_TLS_ALPN, rtpe_config.mqtt_tls_alpn);
                                           ^~~~~~~~~~~~~~~~~
                                           MOSQ_OPT_SSL_CTX
mqtt.c:65:43: note: each undeclared identifier is reported only once for each function it appears in
```

linuxmaniac:
I think this was introduced in v1.6 https://github.com/eclipse/mosquitto/commit/c011be62a41aea8b1e08cb323eb2a59dd36b7add

linuxmaniac:
maybe something like this?
```
diff --git a/daemon/mqtt.c b/daemon/mqtt.c
index 587b179e..869fd82c 100644
--- a/daemon/mqtt.c
+++ b/daemon/mqtt.c
@@ -61,13 +61,15 @@ static int mqtt_connect(void) {
                }
        }
 
-    if (rtpe_config.mqtt_tls_alpn) {
+       #ifdef MOSQ_OPT_TLS_ALPN
+       if (rtpe_config.mqtt_tls_alpn) {
                int ret = mosquitto_string_option(mosq, MOSQ_OPT_TLS_ALPN, rtpe_config.mqtt_tls_alpn);
                if (ret != MOSQ_ERR_SUCCESS) {
                        ilog(LOG_ERR, "Failed to set mosquitto TLS ALPN options: %s", mosquitto_strerror(errno));
                        return -1;
                }
-    }
+       }
+       #endif
 
        ret = mosquitto_connect(mosq, rtpe_config.mqtt_host, rtpe_config.mqtt_port,
                        rtpe_config.mqtt_keepalive);
```

agranig:
Hey! I can check for mosquitto version number, something like

# if LIBMOSQUITTO_VERSION_NUMBER >= 1002000
...
# endif

Let me double-check when this functionality had been introduced in mosquitto.

agranig:
https://github.com/sipwise/rtpengine/pull/1547

linuxmaniac:
Thanks!

PR #1558, opened 2022-10-26
Send metadata mid-call

smititelu:
If metafile connection already exists, send metadata on it, mid-call.


smititelu:
no need to return there, since will reach the end of function anyway

rfuchs:
Doesn't that screw up the contents of the connection? It's supposed to be metadata + nul byte + PCM. Writing the metadata into the middle of the PCM output would leave it indistinguishable from the PCM, no?

smititelu:
Hi,

In my tests, I was able to receive the full mid-call metadata in one read() operation; and I distinguish it by its format (e.g. json).

Each PCM was also received in one separate read() operation.

Thank you,
Stefan

rfuchs:
Ah this is the SEQPACKET forwarding, not the TLS forwarding... Sorry, got the two mixed up. No problem then.

Issue #1559, opened 2022-10-31
Add `log-name` option


Dup4:
rtpengine supports writing logs to syslog and supports setting `log-facilty` to allow writing to a custom log-facilty

However, the only custom log-facilty in syslog is `local0 - local7`

So when I need to deploy multiple rtpengines on one machine and need them to write logs to different files, there is no other way than to configure different `log-facilty`s and the custom log-facilty is limited to 8, which means I can deploy a maximum of 8.

But if we can add an option, `log-name`, we can pass it into the `log_init()` function, so that the `log-name` configuration controls the prefix of log printing, and with the [filter conditions of syslog](https://www.rsyslog.com/doc/master/configuration/filters.html), we can configure all rtpengine instances with the same `log-facilty`, but write them to different files via `log-name`

The advantage of this is that `log-name` has an unlimited number of options.

https://github.com/sipwise/rtpengine/blob/aab897d62e9e549431b715f76971f5773f5643e8/daemon/main.c#L1026

<img width="840" alt="image" src="https://user-images.githubusercontent.com/38343778/199061655-ddca5c53-0e2d-4987-83e1-1d214a260844.png">

If this proposal works, I can submit a PR.

rfuchs:
Sure, happy to accept such a PR. I suggest to put that option into `lib/auxlib.c` so that the recording daemon can benefit too. Remember to document it in the .pod file.

PR #1560, opened 2022-11-01
feat: add log-name option

Dup4:
related #1559

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -1023,7 +1023,7 @@ static void early_init(void) {
 }
 
 static void init_everything(void) {
-	log_init("rtpengine");
+	log_init(rtpe_common_config_ptr->log_name);
 	log_format(rtpe_config.log_format);
 	recording_fs_init(rtpe_config.spooldir, rtpe_config.rec_method, rtpe_config.rec_format);
 	rtpe_ssl_init();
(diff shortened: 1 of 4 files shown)


lib/auxlib.c
```diff
@@ -300,6 +302,9 @@ void config_load(int *argc, char ***argv, GOptionEntry *app_entries, const char
 
 out:
 	// default common values, if not configured
+	if (rtpe_common_config_ptr->log_name == NULL)
+		rtpe_common_config_ptr->log_name = g_strdup("rtpengine");
+
```
rfuchs, line 307: Suggestion to make things more consistent towards the recording daemon: omit this part ...  (1/3)

daemon/main.c
```diff
@@ -1023,7 +1023,7 @@ static void early_init(void) {
 }
 
 static void init_everything(void) {
-	log_init("rtpengine");
+	log_init(rtpe_common_config_ptr->log_name);
```
rfuchs, line 1026: ... leave this one unchanged ... (2/3)

rfuchs (commented):
... and instead put the logic into `log_init()`: if `log_name` is set, use it, or otherwise use the default from the function argument `handle` (3/3)

Does this make sense? Just a suggestion anyway... If you prefer to leave it as it is, I will merge it.

Dup4:
> ... and instead put the logic into `log_init()`: if `log_name` is set, use it, or otherwise use the default from the function argument `handle` (3/3)
> 
> Does this make sense? Just a suggestion anyway... If you prefer to leave it as it is, I will merge it.

I prefer to leave it as it is.

Because I prefer to keep a function clean, rather than coupled with a global configuration.

PR #1563, opened 2022-11-07
Implement playing from a start position.

enreached:
with this option one can play the file from a give position / frame. might be useful to the community.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2093,6 +2093,10 @@ Media files can be provided through one of these keys:
 
 	Contains an integer. How many times to repeat playback of the media. Default is 1.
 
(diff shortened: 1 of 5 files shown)


rfuchs:
I assume "frame" in this context means the sample number? Would it make sense to support this flag for other play modes (e.g. from the DB) as well?

enreached:
> I assume "frame" in this context means the sample number? Would it make sense to support this flag for other play modes (e.g. from the DB) as well?

that is correct. I implemented this for DB and blob too. thanks

rfuchs:
Merged as bab3c94 and dd2956dc

PR #1577, opened 2022-12-06
added EVS 3GPP references

hyavari:


rfuchs:
Do you happen to have a way to test EVS against some real-life application? Because I don't, so the EVS support is entirely untested.

hyavari:
I compiled it and I hope to test it with some Samsung phones this week. I will keep you posted.

PR #1592, opened 2023-01-13
make install - the man pages

dilyanpalauzov:


dilyanpalauzov:
I am not familiar how the installations under Debian work and why the checks fail.  The changes to Makefiles are supposed to improve the installation of man pages under all Linux flavours.
```
   dh_fixperms
   dh_missing
dh_missing: warning: usr/share/man/man8/rtpengine-recording.8 exists in debian/tmp but is not installed to anywhere (related file: "recording-daemon/rtpengine-recording.8")

dh_missing: warning: usr/share/man/man8/rtpengine.8 exists in debian/tmp but is not installed to anywhere (related file: "daemon/rtpengine.8")
dh_missing: error: missing files, aborting
	While detecting missing files, dh_missing noted some files with a similar name to those
	that were missing.  This error /might/ be resolved by replacing references to the
	missing files with the similarly named ones that dh_missing found - assuming the content
	is identical.

	As an example, you might want to replace:
	 * recording-daemon/rtpengine-recording.8
	with:
	 * usr/share/man/man8/rtpengine-recording.8
	in a file in debian/ or as argument to one of the dh_* tools called from debian/rules.
	(Note it is possible the paths are not used verbatim but instead directories 
	containing or globs matching them are used instead)

	Alternatively, add the missing file to debian/not-installed if it cannot and should not
	be used.
```

debian/ngcp-rtpengine-daemon.install
```diff
@@ -1,3 +1,4 @@
 debian/ngcp-rtpengine-iptables-setup /usr/sbin
 etc/rtpengine.conf /etc/rtpengine/
 usr/bin/rtpengine
+usr/share/man/man8/rtpengine.8
```
guillemj, line 4: This needs to be updated instead in the .manpages fragment file, otherwise that's why we get the conflict.

debian/ngcp-rtpengine-recording-daemon.install
```diff
@@ -1,3 +1,4 @@
 debian/ngcp-rtpengine-recording-nfs-setup /usr/sbin/
 etc/rtpengine-recording.conf /etc/rtpengine/
 usr/bin/rtpengine-recording
+usr/share/man/man8/rtpengine-recording.8
```
guillemj, line 4: Same here.

guillemj (changes requested):

debian/ngcp-rtpengine-daemon.install
dilyanpalauzov: ?!? I have not updated “.manpages fragment file”.

Please apply the changes yourself, as I have no idea how installation under Debian is supposed to work.

PR #1599, opened 2023-01-19
Compatibility fixes for SDP session version randomization

lassebm:
- Limit randomized SDP session version to 32-bit integer - many of the SIP stacks use 32-bit integers internally and the Asterisk chan_sip SDP decoder will cap version to 9223372036854775807.
- Avoid randomizing SDP session version at zero - zero is a valid version for new sessions (mentioned in RFC 2327) and still used by some SIP implementations.

If merged, a backport to LTS would be highly appreciated. Thanks.


lassebm:
@rfuchs Can I ask you to backport to 9.5 as well?

PR #1627, opened 2023-03-10
Add an option to not send dtmf log events for injected dtmf

tombriden:
For cases where the same application is receiving DTMF over udp or tcp-ng, and also injecting DTMF, this option allows those injected DTMF to not be sent back on the udp/tcp-ng socket(s).


rfuchs:
LGTM. Please mention the new option in the .pod file and also in the stock/sample config file in etc/

PR #1640, opened 2023-04-04
Fix checksum for kernel packets

smititelu:
Noticed ip and udp checksum problems with recorded RTP packets. Managed to fix checksum only for packets that are from kernel (this PR). What do you think of this patch?

I have a problem with initial packets coming from userspace (e.g. the first 2 RTP packets). They have checksum 0 (invalid checksum) and are using the ```packet->buf``` not ```packet->skbuf```. Do you have an idea how could I fix checksum for those packets too?

I'm looking at this part of the kernel module code:
```
        if (packet->buflen) {
                ret = packet->buflen;
                to_copy = packet->buf;
                DBG("packet is from userspace, %i bytes\n", (int) ret);
        }
        else if (packet->skbuf) {
                ret = packet->skbuf->len;
                to_copy = packet->skbuf->data;
                DBG("packet is from kernel, %i bytes\n", (int) ret);
        }

```

Thank you,
Stefan

### DIFF
--- a/kernel-module/xt_RTPENGINE.c
+++ b/kernel-module/xt_RTPENGINE.c
@@ -3236,6 +3236,10 @@ static ssize_t proc_stream_read(struct file *f, char __user *b, size_t l, loff_t
 	struct re_stream_packet *packet;
 	ssize_t ret;
 	const char *to_copy;
+	struct udphdr *uh;
+	struct iphdr *ih;
+	struct ipv6hdr *ih6;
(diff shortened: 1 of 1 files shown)


smititelu:
Thank you for the advice. Updated PR and tested with ipv4 and ipv6 and it works.

P.S. I am using proc recording.

rfuchs:
Merged as 3397d6e3a

Issue #1641, opened 2023-04-06
Redis restore glib segfault


smititelu:
Hello, I have issues with redis restore when rtpengine is starting and there is some data in redis. Noticed that the issues started since last week, 31 March 2023. Here is the coredump I get:
```
DEBUG: [core] log dtmf over ng 0
DEBUG: [core] no log injected dtmf 0
INFO: [core] Setting timeout for Redis commands to 10 milliseconds
INFO: [core] Asking Redis whether it's master or slave...
INFO: [core] Connected to Redis x.x.x.x:6379 in master mode
INFO: [core] Established initial connection to Redis x.x.x.x6379
INFO: [core] Setting timeout for Redis commands to 10 milliseconds
INFO: [core] Asking Redis whether it's master or slave...
INFO: [core] Connected to Redis x.x.x.x:6379 in slave mode
INFO: [core] Established initial connection to Redis x.x.x.x:6379
INFO: [core] Startup complete, version 1.7.0.jenkins.20230406.1286
[New Thread 0x7ffff5f02700 (LWP 2358)]
[New Thread 0x7ffff5d01700 (LWP 2359)]
DEBUG: [core] timer run time = 0.000014 sec
[New Thread 0x7ffff5b00700 (LWP 2360)]
DEBUG: [core] Restoring calls from Redis...
[Thread 0x7ffff5b00700 (LWP 2360) exited]
INFO: [core] Established initial connection to Redis x.x.x.x:6379
INFO: [core] Established initial connection to Redis x.x.x.x:6379
INFO: [core] Established initial connection to Redis x.x.x.x:6379
INFO: [core] Established initial connection to Redis x.x.x.x:6379
[New Thread 0x7ffff58ff700 (LWP 2361)]
[New Thread 0x7ffff50fe700 (LWP 2362)]
[New Thread 0x7ffff48fd700 (LWP 2363)]
[New Thread 0x7fffe7fff700 (LWP 2364)]

Thread 1 "rtpengine" received signal SIGSEGV, Segmentation fault.
0x00007ffff7db8f26 in g_slice_alloc () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
(gdb) bt
#0  0x00007ffff7db8f26 in g_slice_alloc () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x00007ffff7d95fc6 in g_list_prepend () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
[...]

PR #1642, opened 2023-04-07
Possible fix for issue #1641

smititelu:


rfuchs:
No that's not right, that function call is needed. But I think I see the problem.

PR #1665, opened 2023-05-15
recording-daemon: Insert silence frames as needed during TLS forwarding

tombriden:
If the forwarded streams of a call are later combined without the inserted silence, the resulting mix may end up with the channels out of sync. This is already handled when outputting mixed files so the same functionality from mix.c has been added for tls forwarding

The same thing could arguably also be applied to the single channel output, in which case some refactoring may be better to split common bits out into other functions rather than a mostly copy/pasted function for inserting the silence based on the output type. However, if you're going to mix the single channel outputs then you'd probably just use output-mixed anyway


rfuchs:
Merged as a0fed3781

PR #1675, opened 2023-06-08
recording-daemon : notify-purge param removes a file if notify success

sewan-aprodhomme:
Hi,

We wanted rtpengine to remove records when a notify was successful. We implemented it.  
When a POST request success, we remove the output file if notify-purge option is enabled.

### DIFF
--- a/recording-daemon/main.c
+++ b/recording-daemon/main.c
@@ -62,6 +62,7 @@ int tls_resample = 8000;
 char *notify_uri;
 int notify_post;
 int notify_nverify;
+int notify_purge;
 int notify_threads = 5;
 int notify_retries = 10;
 int notify_record;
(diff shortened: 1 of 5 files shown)


recording-daemon/main.c
```diff
@@ -220,6 +221,7 @@ static void options(int *argc, char ***argv) {
 		{ "notify-no-verify", 	0,   0, G_OPTION_ARG_NONE,	&notify_nverify,"Don't verify HTTPS peer certificate",	NULL		},
 		{ "notify-concurrency",	0,   0, G_OPTION_ARG_INT,	&notify_threads,"How many simultaneous requests",	"INT"		},
 		{ "notify-retries",	0,   0, G_OPTION_ARG_INT,	&notify_retries,"How many times to retry failed requesets","INT"	},
+		{ "notify-purge", 	0,   0, G_OPTION_ARG_STRING,	&notify_purge,	"Remove the local file if notify success", NULL		},
```
rfuchs, line 224: should be _ARG_NONE ?

recording-daemon/notify.c
```diff
@@ -125,6 +125,14 @@ static void do_notify(void *p, void *u) {
 	/* success */
 
 	ilog(LOG_NOTICE, "HTTP notification for '%s%s%s' was successful", FMT_M(req->name));
+
+	if (notify_purge) {
+		if (remove(req->full_filename_path) == 0) {
```
rfuchs, line 130: I think `unlink()` would be more appropriate

rfuchs (commented):
I suppose this option only makes sense if `notify-record` is also set? Perhaps add a safeguard to protect against wrong usage, and/or mention it in the docs?

PR #1678, opened 2023-06-14
Fix for audio stuttering (inserted silence frames)

oeboema:
Hi Richard,

Here's my fix for an issue we had with rtpengine and rtpengine-recording. I saw that the wav files produced by rtpengine-recording had silence frames inserted. The pcaps made by rtpengine where fine. The strange thing was that this happened even with a very small load. I've tested with only one concurrent call, in series with 5 calls of each 2 minutes each. Usually 2 out of 5 calls had missing frames. I found out that the kernel module reported that the queue was full.
This can (in my opinion) only be caused by th recording-daemon not reading fast enough. I made a small modification, now the file descriptor is tested after a frame is read if there is more data available (with a poll()). If so, the next frame is also processed.
This was the queue does not fill up anymore.

### DIFF
--- a/kernel-module/xt_RTPENGINE.c
+++ b/kernel-module/xt_RTPENGINE.c
@@ -3480,7 +3480,7 @@ static void add_stream_packet(struct re_stream *stream, struct re_stream_packet 
 
 	/* discard older packets */
 	while (stream->list_count > stream->info.max_packets) {
-		DBG("discarding old packet from queue\n");
+        log_err("Queue is full, discarding old packet from queue");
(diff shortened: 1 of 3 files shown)


rfuchs:
We have a similar fix on internal review already, one that doesn't need an extra call to poll(). It will be merged in the next few days or so.

rfuchs:
Merged as 5d754605384b2c34222c34d7bd0cf9cfeeedc9b3 and 670f117bcb34a3c3ab9b8e759a4d3b7d97570d8e

Issue #1680, opened 2023-06-19
Crash due to empty learned_endpoint


smititelu:
Hi, we recently got a rtpengine crash in prod:
```
(gdb) bt
#0  0x000055cdd40500c9 in __re_address_translate_ep (ep=0x7f6a685eecd0, o=0x7f6aa67ef818) at media_socket.c:1332
#1  kernelize_one (reti=reti@entry=0x7f6aa67ef800, outputs=outputs@entry=0x7f6aa67ef7e0, stream=stream@entry=0x7f6a685eeac0, sink_handler=0x7f68dc9dffa0, sinks=sinks@entry=0x7f6a685eebb0, 
    payload_types=payload_types@entry=0x7f6aa67ef7d8) at media_socket.c:1418
#2  0x000055cdd40505db in kernelize (stream=0x7f6a685eeac0) at media_socket.c:1626
#3  0x000055cdd4050a0b in media_packet_kernel_check (phc=phc@entry=0x7f6aa67efd90) at media_socket.c:2486
#4  0x000055cdd4050eb2 in stream_packet (phc=phc@entry=0x7f6aa67efd90) at media_socket.c:2652
#5  0x000055cdd4053472 in stream_fd_readable (fd=6306, p=0x7f6ac40ce610, u=<optimized out>) at media_socket.c:3029
#6  0x000055cdd3fef4e8 in poller_poll (p=0x55cdd506b6f0, timeout=<optimized out>) at poller.c:440
#7  0x000055cdd3fefc9d in poller_loop2 (d=0x55cdd506b6f0) at poller.c:619
#8  0x000055cdd3fefe6c in thread_detach_func (d=0x55cdd597cb20) at aux.c:252
#9  0x00007f6b08f96ea7 in start_thread (arg=<optimized out>) at pthread_create.c:477
#10 0x00007f6b08eb6a2f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
``` 

Further debugging looked like it crashed because there was empty "learned_endpoint":
```
...in frame0...
(gdb) print ep
$34 = (const endpoint_t *) 0x7f6a685eecd0
(gdb) print o
$35 = (struct re_address *) 0x7f6aa67ef818

...in frame1...
(gdb) print &reti->expected_src
$38 = (struct re_address *) 0x7f6aa67ef818
(gdb) print &stream->learned_endpoint
$39 = (struct endpoint *) 0x7f6a685eecd0
[...]

Issue #1679, opened 2023-06-16
Drops packets, after re-invite sdp port changed, only when using "asymmetric+strict-source" flags together


smititelu:
Hi, we noticed an issue with RTPEngine dropping RTP packets after re-invite SDP port changed, only when using "asymmetric+strict-source" flags together. Traced back issue starts happening after this commit: 900d2bebee0dfee99c3a10f0391f667b2c93cd01.

RTPEngine logs:
```
[core] Drop due to strict-source attribute; got X.X.X.X:60066, expected X.X.X.X:60064
```
... and packets come in rtpengine but do not get forwarded, for the new 

A fix for this issue would be either:
- don't use "asymmetric" anymore but keep using "strict-source"
- don't use that commit, but keep using "asymmetric+strict-source" flags

I raised this issue, maybe that new commit has some bugs? Please have a look.

Thank you,
Stefan

smititelu:
Full rtpengine logs for the issue:
```
[control] Received command 'offer' from 172.19.92.53:42815
[control] Replying to 'offer' from 172.19.92.53:42815 (elapsed time 0.001055 sec)
[control] Received command 'answer' from 172.19.92.53:35846
[control] Replying to 'answer' from 172.19.92.53:35846 (elapsed time 0.002622 sec)
[core] RTP packet with unknown payload type 13 received from XX.XX.XX.XX:60064
[core] Kernelizing media stream: XX.XX.XX.XX:60064 -> YY.YY.YY.YY:51562 | YY.YY.YY.YY:36662 -> ZZ.ZZ.ZZ.ZZ:42590
[core] Kernelizing media stream: ZZ.ZZ.ZZ.ZZ:42590 -> YY.YY.YY.YY:36662 | YY.YY.YY.YY:51562 -> XX.XX.XX.XX:60064
[control] Received command 'offer' from 172.19.92.53:59469
[core] Removing media stream from kernel: local YY.YY.YY.YY:36662 (dialogue signalling event)
[core] Removing media stream from kernel: local YY.YY.YY.YY:51562 (dialogue signalling event)
[control] Replying to 'offer' from 172.19.92.53:59469 (elapsed time 0.000424 sec)
[core] Kernelizing media stream: ZZ.ZZ.ZZ.ZZ:42590 -> YY.YY.YY.YY:36662 | YY.YY.YY.YY:51562 -> XX.XX.XX.XX:60064
[...]

PR #1682, opened 2023-06-21
fix learning address for asymmetric rtp

emvondo:
fixes for #1679 and #1680

### DIFF
--- a/daemon/media_socket.c
+++ b/daemon/media_socket.c
@@ -1457,7 +1457,7 @@ static const char *kernelize_one(struct rtpengine_target_info *reti, GQueue *out
 
 	if (PS_ISSET2(stream, STRICT_SOURCE, MEDIA_HANDOVER)) {
 		mutex_lock(&stream->out_lock);
-		__re_address_translate_ep(&reti->expected_src, MEDIA_ISSET(media, ASYMMETRIC) ? &stream->learned_endpoint : &stream->endpoint);
+		__re_address_translate_ep(&reti->expected_src, MEDIA_ISSET(media, ASYMMETRIC) && stream->learned_endpoint.address.family ? &stream->learned_endpoint : &stream->endpoint);
 		mutex_unlock(&stream->out_lock);
 		if (PS_ISSET(stream, STRICT_SOURCE))
 			reti->src_mismatch = MSM_DROP;
@@ -1880,13 +1880,13 @@ void __reset_sink_handlers(struct packet_stream *ps) {
 }
 void __stream_unconfirm(struct packet_stream *ps, const char *reason) {
 	__unkernelize(ps, reason);
-	if (!MEDIA_ISSET(ps->media, ASYMMETRIC)) {
-		if (ps->selected_sfd)
-			ilog(LOG_DEBUG | LOG_FLAG_LIMIT, "Unconfirming peer address for local %s (%s)",
-					endpoint_print_buf(&ps->selected_sfd->socket.local),
-					reason);
+
+	if (ps->selected_sfd)
+		ilog(LOG_DEBUG | LOG_FLAG_LIMIT, "Unconfirming peer address for local %s (%s)",
+			endpoint_print_buf(&ps->selected_sfd->socket.local),
+			reason);
 		PS_CLEAR(ps, CONFIRMED);
-	}
+
 	__reset_sink_handlers(ps);
 }
 static void stream_unconfirm(struct packet_stream *ps, const char *reason) {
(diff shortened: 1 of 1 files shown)


daemon/media_socket.c
```diff
@@ -2515,10 +2508,12 @@ static bool media_packet_address_check(struct packet_handler_ctx *phc)
...
 	{
-		endpoint = phc->mp.stream->endpoint;
-		phc->mp.stream->endpoint = *use_endpoint_confirm;
+		endpoint = !MEDIA_ISSET(phc->mp.media, ASYMMETRIC) ? phc->mp.stream->endpoint : phc->mp.stream->learned_endpoint;
+		if (!MEDIA_ISSET(phc->mp.media, ASYMMETRIC))
+			phc->mp.stream->endpoint = *use_endpoint_confirm;
+
 		phc->mp.stream->learned_endpoint = *use_endpoint_confirm;
-		if (memcmp(&endpoint, &phc->mp.stream->endpoint, sizeof(endpoint))) {
+		if (memcmp(&endpoint, !MEDIA_ISSET(phc->mp.media, ASYMMETRIC) ? &phc->mp.stream->endpoint : &phc->mp.stream->learned_endpoint , sizeof(endpoint))) {
```
rfuchs, line 2516: All these 3 case distinctions are based on the same condition, so you could reduce this to just one if/else.

In particular, if ASYMMETRIC is set, then the final memcmp() condition will never be true, because it ends up comparing `learned_endpoint` against `endpoint`, and `endpoint` has just been set to `learned_endpoint` 3 lines earlier, with no chance of it ending up something different, no?

rfuchs (commented):
Still not quite sure about the logic here 😅

daemon/media_socket.c
emvondo: do you mean something like

```
if (MEDIA_ISSET(phc->mp.media, ASYMMETRIC)){
    endpoint = phc->mp.stream->learned_endpoint;

}else{
    endpoint = phc->mp.stream->endpoint;
    phc->mp.stream->endpoint = *use_endpoint_confirm;
}

phc->mp.stream->learned_endpoint = *use_endpoint_confirm;
```

> In particular, if ASYMMETRIC is set, then the final memcmp() condition will never be true,

yes but isn't that a specific case ? the same logic applies if ASYMMETRIC is not set we will end up comparing stream->endpoint to endpoint (set to stream->endpoint earlier). we are comparing old value with current value.

> with no chance of it ending up something different, no?

yes if the value does not change but the same logic applies for non ASYMMETRIC  stream i think ?

daemon/media_socket.c
```diff
@@ -2396,15 +2396,8 @@ static bool media_packet_address_check(struct packet_handler_ctx *phc)
 
 	PS_SET(phc->mp.stream, RECEIVED);
 
-	/* do not pay attention to source addresses of incoming packets for asymmetric streams */
-	if (MEDIA_ISSET(phc->mp.media, ASYMMETRIC) || phc->mp.stream->el_flags == EL_OFF) {
+	if (phc->mp.stream->el_flags == EL_OFF || rtpe_config.endpoint_learning == EL_OFF)
```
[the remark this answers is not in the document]
emvondo: yes i agree rtpe_config.endpoint_learning check is not necessary

daemon/media_socket.c
rfuchs: > do you mean something like
> 
> ```
> if (MEDIA_ISSET(phc->mp.media, ASYMMETRIC)){
>     endpoint = phc->mp.stream->learned_endpoint;
> 
> }else{
>     endpoint = phc->mp.stream->endpoint;
>     phc->mp.stream->endpoint = *use_endpoint_confirm;
> }
> 
> phc->mp.stream->learned_endpoint = *use_endpoint_confirm;
> ```

Yes exactly.

> 
> > In particular, if ASYMMETRIC is set, then the final memcmp() condition will never be true,
> 
> yes but isn't that a specific case ? the same logic applies if ASYMMETRIC is not set we will end up comparing stream->endpoint to endpoint (set to stream->endpoint earlier). we are comparing old value with current value.
> 
> > with no chance of it ending up something different, no?
> 
> yes if the value does not change but it could change (especially for asymmetric) but the same logic applies for non ASYMMETRIC stream i think ?

I don't see how it could change. Consider your code above, with MEDIA_ISSET(phc->mp.media, ASYMMETRIC) being true. You set `endpoint` to `learned_endpoint`. And then immediately after you compare `endpoint` against `learned_endpoint`. How could these two ever be different?

For the non-asymmetric case you compare the previously learned endpoint (saved in `endpoint`) with the newly learned endpoint (`use_endpoint_confirm` stored into `phc->mp.stream->endpoint`) and use that to trigger a log message. (Not that the log message is crucial, but it makes me think that the logic isn't sound)

emvondo: > I don't see how it could change. Consider your code above, with MEDIA_ISSET(phc->mp.media, ASYMMETRIC) being true. You set endpoint to learned_endpoint. And then immediately after you compare endpoint against learned_endpoint. How could these two ever be different?

because before memcmp we update **learned_endpoint**  

phc->mp.stream->learned_endpoint = *use_endpoint_confirm;

emvondo: hum ok sorry made a mistake. i did not see im making the comparison twice.
i will update it

rfuchs:
stale

PR #1685, opened 2023-06-22
Replace bash shebangs with sh

orgads:
There are no bash features in these scripts, and some distributions don't install bash by default.

### DIFF
--- a/utils/build_test_wrapper
+++ b/utils/build_test_wrapper
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 TARGET=$1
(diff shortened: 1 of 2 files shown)


rfuchs:
Have you checked whether any of these scripts contain any bashisms?

rfuchs:
Shellcheck is not happy with this: https://github.com/sipwise/rtpengine/actions/runs/5342298437/jobs/9690717029?pr=1685

orgads:
Nice, thanks!

Reverted the broken scripts.

PR #1729, opened 2023-09-27
Add sysfs parameters

IgorrG:
Sometimes it is useful to check actual module parameters.


kernel-module/xt_RTPENGINE.c
```diff
@@ -195,29 +195,29 @@ struct rtpengine_output;
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(3,10,0)
 static kuid_t proc_kuid;
 static uint proc_uid = 0;
-module_param(proc_uid, uint, 0);
+module_param(proc_uid, uint, S_IRUGO | S_IWUSR);
```
rfuchs, line 198: The modparams for the /proc entries are only used during module load, so there isn't really any point in having them writable (unless you add your own setter function)

kernel-module/xt_RTPENGINE.c
```diff
@@ -195,29 +195,29 @@ struct rtpengine_output;
...
 #else
-module_param(proc_mask, uint, 0);
+module_param(proc_mask, uint, S_IRUGO | S_IWUSR);
 #endif
 MODULE_PARM_DESC(proc_mask, "rtpengine procfs tree mode mask");
 #endif
 
 static uint stream_packets_list_limit = 10;
-module_param(stream_packets_list_limit, uint, 0);
+module_param(stream_packets_list_limit, uint, S_IRUGO | S_IWUSR);
```
rfuchs, line 216: This may have undesired effects if set to zero (and is only checked during module load). OK to leave it like this for now, but may want to have a setter function for this in the future.

rfuchs (commented):
LGTM for the most part. Let me know what you think about my comments.

kernel-module/xt_RTPENGINE.c
IgorrG: Agree and actually do not see currently a reason to make parameters writable, only `log_errors` make sense to be writeable.

kernel-module/xt_RTPENGINE.c
IgorrG: Much better to leave it read only for now.

PR #1736, opened 2023-10-11
Recording without sinks

john08burke:
This PR allows for the `recording-daemon` to process media without any sinks. One such use-case would be to allow for dedicated recording-only media servers. An implementation flow via rtpengine only media servers would be as follows:

1. **subscribe request** to Proxy A
2. **publish** to Proxy B
3. **subscribe answer** to Proxy A
4.  **start recording** to Proxy B

I did run into an issue with the `DBG` func compiling, so I mirrored the logs found on `master`.

### DIFF
--- a/kernel-module/xt_RTPENGINE.c
+++ b/kernel-module/xt_RTPENGINE.c
@@ -4506,7 +4506,7 @@ static unsigned int rtpengine46(struct sk_buff *skb, struct rtpengine_table *t, 
 	}
 	_r_unlock(&g->outputs_lock, flags);
 
-	DBG("target found, src "MIPF" -> dst "MIPF"\n", MIPP(g->target.src_addr), MIPP(g->target.dst_addr));
+	DBG("target found, local " MIPF "\n", MIPP(g->target.local));
 	DBG("target decrypt hmac and cipher are %s and %s", g->decrypt.hmac->name,
 			g->decrypt.cipher->name);
 
@@ -4543,11 +4543,6 @@ not_stun:
 src_check_ok:
 	if (g->target.dtls && is_dtls(skb))
 		goto skip1;
-	if (g->target.non_forwarding) {
-		if (g->target.blackhole)
-			error_nf_action = NF_DROP;
-		goto skip1;
-	}
 
 	rtp.ok = 0;
 	if (!g->target.rtp)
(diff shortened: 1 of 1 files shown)


kernel-module/xt_RTPENGINE.c
```diff
@@ -4544,6 +4544,8 @@ static unsigned int rtpengine46(struct sk_buff *skb, struct rtpengine_table *t,
 	if (g->target.dtls && is_dtls(skb))
 		goto skip1;
 	if (g->target.non_forwarding) {
+		if (g->target.do_intercept)
+			goto not_rtp;
 		if (g->target.blackhole)
 			error_nf_action = NF_DROP;
 		goto skip1;
```
rfuchs, line 4551: I think it would be more appropriate to defuse this entire block when `do_intercept` is set, instead of jumping to `not_rtp`. The `not_rtp` label is specifically meant for media that isn't RTP, and having interception active doesn't mean it's not RTP. For the SRTP case for example you'd still want it to be decrypted.

So I think the appropriate code path with both `non_forwarding` and `blackhole` set would be to still go through almost everything below (including decryption and intercept) and then finally just not forward anything and drop the packet

john08burke: Good points. I just pushed a revised commit which hopefully addresses your comments.

I also noticed the stats weren't incrementing on `non_forwarding` and `blackhole` scenarios. By delaying the drop logic to the final step it also solves this problem.

```
cat /proc/rtpengine/0/list
local inet4 192.168.26.246:10000
    stats:               107156 bytes,                  623 packets,                    0 errors
    SSRC in: f915192d
    option: non forwarding
    option: blackhole
local inet4 192.168.26.246:10016
    stats:               107328 bytes,                  624 packets,                    0 errors
    SSRC in: f930ca98
    option: non forwarding
    option: blackhole

```

john08burke:
Ha... I completely missed that when I looked through master, but yes that's pretty much the same solution path. Can we backport that to 10.5 (we have yet to do the upgrade to 11)?

john08burke:
We can close this one, as you've already fixed it. I will manually backport on my local repo and have plans on upgrading to 11.5 soon!

Thanks for the help with this one and sorry for missing the patch on your later releases!

PR #1753, opened 2023-11-13
RPM packaging: update for EL8/EL9

space88man:
Update RPM packaging for EL8/EL9
- add example mock chroots
- update for nftables, iptables deprecation in EL9
- EL8: use newer GCC

### DIFF
--- a/el/rtpengine.spec
+++ b/el/rtpengine.spec
@@ -14,7 +14,7 @@ Conflicts:	%{name}-kernel < %{version}-%{release}
 BuildRequires:	gcc make pkgconfig redhat-rpm-config
 BuildRequires:	glib2-devel libcurl-devel openssl-devel pcre-devel
 BuildRequires:	xmlrpc-c-devel zlib-devel hiredis-devel
-BuildRequires:	libpcap-devel libevent-devel json-glib-devel 
+BuildRequires:	libpcap-devel libevent-devel json-glib-devel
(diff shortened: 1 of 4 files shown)


rfuchs:
@mika Any opinions on this?

el/rtpengine.spec
```diff
@@ -236,15 +249,25 @@ true
 %attr(0750,%{name},%{name}) %dir %{_var}/spool/%{binname}
 # Documentation
 %doc LICENSE README.md debian/changelog debian/copyright
+%{_mandir}/man8/%{binname}.8*
 
 %files kernel
-/%{_lib}/xtables/libxt_RTPENGINE.so
+## /usr/lib64/xtables/libxt_RTPENGINE.so
```
mika, line 255: Hm, what's the idea behind this change?

mika (commented):
LGTM overall (unverified/tested though), only one comment about `libxt_RTPENGINE.so`

el/rtpengine.spec
space88man: That artifact  was removed in eea3824725

mika: Right, but shouldn't we _drop_ it then here? :)

space88man: Done and pushed

mika (approved):
Untested, but LGTM, thanks @space88man

Issue #1754, opened 2023-11-14
[build] Use of automatic variable in prerequisite list  of lib/common.Makefile rule is ineffective


space88man:
Problem: compilation observed during `make install` step after a successful `make` .

This is unexpected and some artifacts were being rebuilt. Tracked this down to the use of an automatic variable in a prerequisite which is a mistake according to the GNU Make docs.

```
# in lib/common.Makefile...
$(BUILD_TEST_ALTS): ../lib/$(@:.h=-*)
        ../utils/build_test_wrapper "$@" 2> /dev/null
```

This is ineffective according to **https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html**

> Also, they cannot be accessed directly within the prerequisite list of a rule. A common mistake is attempting to use $@ within the prerequisites list; this will not work.

The expansion is empty and the rule will always be run. E.g. `daemon/fix_frame_channel_layout.h` was considered older than `lib/` (note: bare directory only and not the actual prerequisites `lib/fix_frame_channel-*`).

Proposed fix:
```
# Either 
.SECONDEXPANSION:
$(BUILD_TEST_ALTS): ../lib/$$(subst .h,-*,$$@)
        ../utils/build_test_wrapper "$@" 2> /dev/null

# or
.SECONDEXPANSION:
$(BUILD_TEST_ALTS):  %.h : $$(wildcard ../lib/%-*)
        ../utils/build_test_wrapper "$@" 2> /dev/null

```

Update: it seems that BSD make supports this use of `$@` to derive dependencies from this post https://forums.freebsd.org/threads/posix-compliant-makefile-pattern-rules.65372/post-438485.

When I tried the existing Makefile with GNU make the dependency is always output as `lib/`.

PR #1755, opened 2023-11-14
Fix lib/common.Makefile use of automatic variable in prerequisite list

space88man:
Replace ineffective use of automatic variable in prerequisite list with a pattern rule.

Addresses #1754


rfuchs:
@guillemj  what's your take on this?

rfuchs:
Already fixed

PR #1759, opened 2023-11-22
Add RTCP generation interval to config

andingv:
This pull request makes the RTCP generation interval that gets used, when generate-RTC=on is passed for offer, configurable, rather than hardcoded in the daemon.


andingv:
Hi, I added it, please check

PR #1769, opened 2023-12-05
Avoiding "Too many packets in UDP receive queue (more than 50), aborting loop. Dropped packets possible" errors.

Den4t:
Hi !

This fixation avoids the errors "Too many packets in the UDP receive queue (more than 50) interrupting the loop. Dropped packets are possible" by moving to a configurable variable counter of continuous read cycles (new config parameter max_recv_iters), and blocking of stream read operations after this error occurred has also been fixed.

### DIFF
--- a/daemon/main.c
+++ b/daemon/main.c
@@ -106,6 +106,7 @@ struct rtpengine_config rtpe_config = {
 			[log_level_index_internals] = -1,
 		},
 	},
+	.max_recv_iters = MAX_RECV_ITERS,
 };
 
 static void sighandler(gpointer x) {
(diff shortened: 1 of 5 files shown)


daemon/main.c
```diff
@@ -789,6 +791,9 @@ static void options(int *argc, char ***argv) {
 	if (rtpe_config.control_tos < 0 || rtpe_config.control_tos > 255)
 		die("Invalid control-ng TOS value");
 
+	if (rtpe_config.max_recv_iters < MAX_RECV_ITERS || rtpe_config.control_tos > MAX_RECV_ITERS*10)
```
rfuchs, line 794: copy/paste error (left `control_tos`)

I don't think there's a reason to bound-check this value at all, except maybe to make sure it's not negative.

rfuchs (changes requested):
Also please document the new option in `docs/rtpengine.md`

daemon/main.c
Den4t: I think the value should be at least 1, otherwise the error occurs at the first iteration of the loop.

Den4t:
> Also please document the new option in `docs/rtpengine.md`

Done.
> I don't think there's a reason to bound-check this value at all, except maybe to make sure it's not negative.
Also.

PR #1772, opened 2023-12-11
reuse the master key for SRTP when DTLS restarts

hajekmi:
Call setup:
1. WebRTC client (DTLS-SRTP)
2. Kam + RTPEngine latest
3. RTPEngine decrypt and send to Asterisk over SIP + RTP/RTCP
4. WebRTC client change IP address during the call
5. Client restart ICE and send renegotiate (new handshake DTLS and Re-INVITE)
6. The master key (SRTP) remains the same (on Chrome, Firefox, Safari)
7. WebRTC client send SRTP to RTPEngine with same master key for SRTP
8. RTPEngine have issue with  encrypt and decrypt this S/RTP (from WebRTC and from Asterisk)

Issue is in the RTPEngine, when restart DTLS make clean Master key for SRTP.
This behaviour has been addressed here in Chromium https://bugs.chromium.org/p/webrtc/issues/detail?id=5205
I attached fix.

### DIFF
--- a/daemon/dtls.c
+++ b/daemon/dtls.c
@@ -912,10 +912,7 @@ void dtls_shutdown(struct packet_stream *ps) {
 		}
 
 		dtls_connection_cleanup(d);
-
-		crypto_reset(&sfd->crypto);
 	}
-
 
 	if (ps->dtls_cert) {
 		X509_free(ps->dtls_cert);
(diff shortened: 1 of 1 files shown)


rfuchs:
Would line 916/917 also need to be removed? https://github.com/sipwise/rtpengine/pull/1772/files#diff-d701009e80a993f1a5f7309566fe3e0bd6dd2f7be97a1aa2cc36f8b93ed674c1R916

`sfd->crypto` holds the SRTP context for ingress media

PR #1777, opened 2023-12-21
Add option to use pure TCP socket to forward call data (used to reduc…

IgorrG:
…e overal system load on local audio transmition).

### DIFF
--- a/recording-daemon/packet.c
+++ b/recording-daemon/packet.c
@@ -113,24 +113,6 @@ void ssrc_tls_state(ssrc_t *ssrc) {
 			dbg("TLS connection to %s doing handshake",
 				endpoint_print_buf(&tls_send_to_ep));
 			ssrc->tls_fwd_poller.state = PS_HANDSHAKE;
-			if (!tls_disable) {
-				if ((ret = SSL_connect(ssrc->ssl)) == 1) {
-					dbg("TLS connection to %s established",
-							endpoint_print_buf(&tls_send_to_ep));
-					ssrc->tls_fwd_poller.state = PS_OPEN;
-					streambuf_writeable(ssrc->tls_fwd_stream);
-				}
-				else
-					ssrc_tls_check_blocked(ssrc->ssl, ret);
-			}
-		}
-		else if (status < 0) {
-			ilog(LOG_ERR, "Failed to connect TLS socket: %s", strerror(errno));
-			ssrc_tls_shutdown(ssrc);
-		}
-	}
-	else if (ssrc->tls_fwd_poller.state == PS_HANDSHAKE) {
-		if (!tls_disable) {
 			if ((ret = SSL_connect(ssrc->ssl)) == 1) {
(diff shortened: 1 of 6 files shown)


recording-daemon/packet.c
```diff
@@ -111,29 +113,35 @@ void ssrc_tls_state(ssrc_t *ssrc) {
 			dbg("TLS connection to %s doing handshake",
 				endpoint_print_buf(&tls_send_to_ep));
 			ssrc->tls_fwd_poller.state = PS_HANDSHAKE;
-			if ((ret = SSL_connect(ssrc->ssl)) == 1) {
-				dbg("TLS connection to %s established",
-						endpoint_print_buf(&tls_send_to_ep));
-				ssrc->tls_fwd_poller.state = PS_OPEN;
-				streambuf_writeable(ssrc->tls_fwd_stream);
+			if (!tls_disable) {
```
rfuchs, line 116: Wouldn't it make sense to set state = PS_OPEN immediately here in the non-TLS case?

rfuchs (commented):
Please document this new option in docs/rtpengine-recording.md

recording-daemon/packet.c
IgorrG: Thanks, make sense.

recording-daemon/packet.c
```diff
@@ -290,12 +284,13 @@ static ssrc_t *ssrc_get(stream_t *stream, unsigned long ssrc) {
 				goto tls_out;
 			}
 			ret->tls_fwd_stream = streambuf_new_ptr(&ret->tls_fwd_poller, ret->ssl, &ssrc_tls_funcs);
+			ssrc_tls_state(ret);
 		} else {
 			ret->tls_fwd_stream = streambuf_new(&ret->tls_fwd_poller, ret->tls_fwd_sock.fd);
+			ret->tls_fwd_poller.state = PS_OPEN;
```
rfuchs, line 290: I think that's a bit too aggressive now 😆

`connect_socket_nb()` doesn't necessarily give you an open socket immediately, so state should still go to PS_CONNECTING first. But in `ssrc_tls_state()`, with state `PS_CONNECTING`, when `connect_socket_retry()` then returns success, state can be set to `PS_OPEN` instead of `PS_HANDSHAKE`.

IgorrG:
Name of option 'tls-disable' maybe confusing, it could be good idea to change it to 'tls-disable-handshake' or 'tls-force-tcp'?

rfuchs:
> Name of option 'tls-disable' maybe confusing, it could be good idea to change it to 'tls-disable-handshake' or 'tls-force-tcp'?

Yes, I agree. I'm actually thinking of renaming the `tls-...` options to `tcp-...` and then have an option to enable TLS, instead of the other way around.

Backwards compatibility with existing config files would be an issue though. Either retain the `tls-...` options as an alternative fallback, and/or warn users when they upgrade and don't change the options.

Issue #1781, opened 2024-01-15
Failure to generate RTCP with requested interval


szcom:
### rtpengine version the issue has been seen with

Version: 11.5.1.11+0~mr11.5.1.11 git-HEAD-51248ea

### CPU architecture issue was seen on (see `uname -m`)

None

### Expected behaviour you didn't see

1. Caller (AMR), answer(g711), rtpengine does transcoding. It is asked to "generate RTCP" in NG offer/answer flags.
2. Rtpengine produces RTCP to both sides
3. Caller puts answerer on-hold, "generate RTCP" is in offer/answer flags
4. Rtpengine keeps producing RTCP

I did not see 4 happening, instead Rtpengine stops sending RTCP and caller drops the call

### Unexpected behaviour you saw

Rtpengine stops generating RTCP when call is put on-hold, caller says a=sendonly and answerer acks it by a=recvonly
"generate RTCP" is also present in offer/answer with sendonly/recvonly

szcom:
rtcp-interval is in fact doubled, this causes RTCP to arrive with >10s intervals. In some situations it is too late. the PR makes RTCP to tick with value configured in rtcp-interval

PR #1783, opened 2024-01-17
Fix RTCP report deadline timeout ref #1781

szcom:


rfuchs:
Is this a fix for #1781 or unrelated?

szcom:
Yes it is the fix for #1781

PR #1787, opened 2024-01-29
Fix issues when nftables-base-chain is "none"

gaaf:
Don't delete or fail on items the admin is expected to setup in this mode of operation.

### DIFF
--- a/daemon/nftables.c
+++ b/daemon/nftables.c
@@ -562,7 +562,7 @@ static const char *nftables_shutdown_family(struct mnl_socket *nl, int family, u
 
 	err = delete_chain(nl, family, seq, chain);
 	if (err) {
-		if (errno != ENOENT) // ignore trying to delete stuff that doesn't exist
+		if (errno != ENOENT && errno != EBUSY) // ignore trying to delete stuff that doesn't exist
 			return err;
 	}
 
--- a/daemon/nftables.c
+++ b/daemon/nftables.c
@@ -518,27 +518,31 @@ static const char *delete_chain(struct mnl_socket *nl, int family, uint32_t *seq
 static const char *nftables_shutdown_family(struct mnl_socket *nl, int family, uint32_t *seq,
 		const char *chain, const char *base_chain, nftables_args *dummy)
 {
-	// clean up rules in legacy `INPUT` chain
-	const char *err = iterate_rules(nl, family, "INPUT", seq,
-			&(struct iterate_callbacks) {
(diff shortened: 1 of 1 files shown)


daemon/nftables.c
```diff
@@ -518,27 +518,31 @@ static const char *delete_chain(struct mnl_socket *nl, int family, uint32_t *seq
...
+	if (!base_chain || strcmp(base_chain, "none")) {
+		// clean up rules in legacy `INPUT` chain
+		err = iterate_rules(nl, family, "INPUT", seq,
+				&(struct iterate_callbacks) {
+					.parse_expr = match_immediate_rtpe,
+					.chain = chain,
+					.rule_final = check_immediate,
+					.iterate_final = iterate_delete_rules,
+				});
+		if (err)
```
dfxdj, line 532: Does this commit fix a particular issue that you're seeing? Because this was done like this on purpose, to make transitioning from the previous iptables-extension package easier and automatic.

gaaf: Yes. I have a custom rule in the INPUT chain to jump to the rtpengine chain. This jump rule is removed at rtpengine shutdown **and startup**.

This change should only affect configs with `nftables-base-chain = none`. The default is not "none", so it should not affect people transitioning.

Documentation for  `nftables-base-chain` says:
> If this option is set to the special string none, then rtpengine will create its custom chain and rule as it normally would, but will skip adding an immediate-goto rule to the custom chain. Doing so requires the operator to manually create this immediate-goto rule somewhere themselves.

This won't work if rtpengine deletes said rule.

rfuchs: > Yes. I have a custom rule in the INPUT chain to jump to the rtpengine chain. This jump rule is removed at rtpengine shutdown **and startup**.
> 
> This change should only affect configs with `nftables-base-chain = none`. The default is not "none", so it should not affect people transitioning.
> 
> Documentation for `nftables-base-chain` says:
> 
> > If this option is set to the special string none, then rtpengine will create its custom chain and rule as it normally would, but will skip adding an immediate-goto rule to the custom chain. Doing so requires the operator to manually create this immediate-goto rule somewhere themselves.
> 
> This won't work if rtpengine deletes said rule. My config:
> 
> ```
> nftables-chain = rtpengine-0
> nftables-base-chain = none
> nftables-family = ip
> ```
> 
> Also, IMHO rtpengine should not touch and/or remove rules and/or chains it did not add itself.

Alright, makes sense, thanks

PR #1802, opened 2024-03-04
Adding support for NG trace to Homer

lbalaceanu:
Proposing a patch to also send NG traffic to sipcapture servers.

For a sipcapture server one can use Kamailio and insert from the config similar to https://github.com/sipcapture/homer-config/blob/master/sipcapture/sipcapture.kamailio5. 

I am setting the protocol type to 0x3d (61) - rtcp homer fills it with 0x05.

Support is desired for Kamailio/Rtpengine traffic via UDP. Adding homer-enable-rtcp-stats and homer-enable-ng config params to separately control sending to Homer each traffic type. By default rtcp is `on` when homer parameter is configured. NG is by default `off`.

Any comments are appreciated,
Thank you

### DIFF
--- a/daemon/control_ng.c
+++ b/daemon/control_ng.c
@@ -20,6 +20,7 @@
 #include "statistics.h"
 #include "streambuf.h"
 #include "str.h"
+#include "homer.h"
 #include "tcp_listener.h"
 #include "main.h"
 
@@ -28,6 +29,7 @@ mutex_t tcp_connections_lock;
 GHashTable *rtpe_cngs_hash;
 GHashTable *tcp_connections_hash;
 static struct cookie_cache ng_cookie_cache;
+static bool trace_ng = false;
 
 const char magic_load_limit_strings[__LOAD_LIMIT_MAX][64] = {
 	[LOAD_LIMIT_MAX_SESSIONS] = "Parallel session limit reached",
@@ -62,6 +64,67 @@ const char *ng_command_strings_short[NGC_COUNT] = {
 	"Pub", "SubReq", "SubAns", "Unsub",
 };
 
+typedef struct ng_ctx {
+	str callid;
+	int command;
+	str cookie;
+	bool should_trace;
+	const endpoint_t *sin_ep;
+	const endpoint_t *local_ep;
+} ng_ctx;
+
+#define CH(func, ...) do { \
+	if (trace_ng) \
+		func( __VA_ARGS__); \
+} while (0)
+
+void init_ng_tracing(void) {
+	if (rtpe_config.homer_ng_on &&  has_homer())
+		trace_ng = true;
+}
+
+static GString *create_homer_msg(str *cookie, str *data) {
+	GString *msg = g_string_sized_new(cookie->len + 1 + data->len);
+	g_string_append_printf(msg, "%.*s %.*s", STR_FMT(cookie), STR_FMT(data));
+	return msg;
+}
+
+static bool should_trace_msg(enum ng_command command) {
+	switch (command) {
+		case NGC_PING:
+			return false;
+		default:
+			return true;
+	}
+}
+
+static void homer_fill_values(ng_ctx *hctx, str *callid, int command) {
+	if (hctx) {
+		hctx->command = command;
+		hctx->callid = *callid;
+	}
+}
+
+static void homer_trace_msg_in(ng_ctx *hctx, str *data) {
+	if (hctx) {
+		hctx->should_trace = should_trace_msg(hctx->command);
+		if (hctx->should_trace)	{
+			struct timeval tv;
+			gettimeofday(&tv, NULL);
+			GString *msg = create_homer_msg(&hctx->cookie, data);
+			homer_send(msg, &hctx->callid, hctx->sin_ep, hctx->local_ep, &tv, rtpe_config.homer_ng_capt_proto);
+		}
+	}
+}
+
+static void homer_trace_msg_out(ng_ctx *hctx, str *data) {
+	if (hctx && hctx->should_trace) {
+		struct timeval tv;
+		gettimeofday(&tv, NULL);
+		GString *msg = create_homer_msg(&hctx->cookie, data);
+		homer_send(msg, &hctx->callid, hctx->local_ep, hctx->sin_ep, &tv, rtpe_config.homer_ng_capt_proto);
+	}
+}
 
 static void pretty_print(bencode_item_t *el, GString *s) {
 	bencode_item_t *chld;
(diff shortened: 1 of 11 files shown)


lbalaceanu:
Hi @rfuchs, do you consider such an addition welcomed?

rfuchs:
> Hi @rfuchs, do you consider such an addition welcomed? Thank you.

Yes absolutely. I'll take a closer look at this when I find some spare time.

I assume the protocol value of 0x3d has been registered on the Homer side, or is a free-form value?

lbalaceanu:
Hi Richard, 

I just picked this 0x3d value as first one free in the HEP3 protocol specification, maybe making the variable configurable is the way to go.

From what I know, there is no official sipcapture/Homer UI support for capturing/displaying NG values. One can still look the data up in the database itself. 

At 1&1 we use an old Homer interface and with little hacks this NG can be displayed alongside the SIP itself just as long as the SIP and the NG messages are stored in the same tables.

Thank you

daemon/control_ng.c
```diff
@@ -62,6 +63,104 @@ const char *ng_command_strings_short[NGC_COUNT] = {
...
+		//"Failed to parse JSON document";
+		if (!json_parser_load_from_data(json, data->s, data->len, NULL))
+			goto end_function;
+		dict = bencode_convert_json(&ngbuf->buffer, json);
+		//"Could not decode bencode dictionary";
+		if (!dict || dict->type != BENCODE_DICTIONARY)
+			goto end_function;
+	}
+	else {
+		//"Invalid NG data format";
```
rfuchs, line 132: Any way to avoid duplicating this code? Is this even needed? AFAICT it's only used for sending cached replies. Cached replies are retransmits by definition. Are these of interest to the Homer backend?

daemon/main.c
```diff
@@ -578,6 +579,8 @@ static void options(int *argc, char ***argv) {
 		{ "homer",	0,  0, G_OPTION_ARG_STRING,	&homerp,	"Address of Homer server for RTCP stats","IP46|HOSTNAME:PORT"},
 		{ "homer-protocol",0,0,G_OPTION_ARG_STRING,	&homerproto,	"Transport protocol for Homer (default udp)",	"udp|tcp"	},
 		{ "homer-id",	0,  0, G_OPTION_ARG_INT,	&rtpe_config.homer_id,	"'Capture ID' to use within the HEP protocol", "INT"	},
+		{ "homer-enable-rtcp-stats", 0, 0, G_OPTION_ARG_INT,	&rtpe_config.homer_rtcp,	"Enable/disable RTCP stats to Homer (enabled by default if homer server enabled, else disabled)", "0|1"	},
+		{ "homer-enable-ng", 0, 0, G_OPTION_ARG_INT,	&rtpe_config.homer_ng,	"Enable/disable NG to Homer", "0|1"	},
```
rfuchs, line 583: I'd say those should be boolean values, but I understand the problem with the RTCP switch. You should be able to work around it by using a `gboolean` and still setting it to -1 (or some other magic value), or perhaps by using a `G_OPTION_ARG_CALLBACK`. Alternatively a negated option meaning would also work.

rfuchs (commented):
A configurable protocol value would be preferable I think, yes.

daemon/main.c
lbalaceanu: Hi, if OK with you I would go for a negated option homer-disable-rtcp-stats and the normal homer-enable-ng, both gboolean.

daemon/control_ng.c
lbalaceanu: Hi, I think that in debugging scenarios one would be interested to see that retransmissions are sent. To avoid duplication of code maybe I can also cache some values just as the reply itself is cached.

daemon/main.c
rfuchs: > Hi, if OK with you I would go for a negated option homer-disable-rtcp-stats and the normal homer-enable-ng, both gboolean.

Sounds good

lbalaceanu:
I tried addressing all but the first comments and I now get an error when the unit tests are run. I am a bit at a loss to identify which scenario detects the problem.

Thank you

rfuchs:
That's libasan complaining. There's a mem leak.

```
==340490==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 18120 byte(s) in 755 object(s) allocated from:
    #0 0x7f97594f3bd7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x56233954ab4a in cache_entry_dup ../include/cookie_cache.h:25
    #2 0x56233954ab4a in cookie_cache_insert /home/dfx/src/ngcp-git/rtpengine/daemon/cookie_cache.c:74
    #3 0x56233956d132 in control_ng_process /home/dfx/src/ngcp-git/rtpengine/daemon/control_ng.c:509
    #4 0x56233956ea7a in control_ng_incoming /home/dfx/src/ngcp-git/rtpengine/daemon/control_ng.c:554
    #5 0x56233954e28a in udp_listener_incoming /home/dfx/src/ngcp-git/rtpengine/daemon/udp_listener.c:56
    #6 0x562339b003bc in poller_poll /home/dfx/src/ngcp-git/rtpengine/daemon/poller.c:224
    #7 0x562339b02e72 in poller_loop /home/dfx/src/ngcp-git/rtpengine/daemon/poller.c:321
    #8 0x56233940ec82 in thread_detach_func /home/dfx/src/ngcp-git/rtpengine/daemon/helpers.c:246
    #9 0x7f975945ae65 in asan_thread_start ../../../../src/libsanitizer/asan/asan_interceptors.cpp:234

SUMMARY: AddressSanitizer: 18120 byte(s) leaked in 755 allocation(s).

```

You can run this locally with `make asan-check`. Don't forget to clean sources between builds.

include/cookie_cache.h
```diff
@@ -12,6 +12,30 @@ struct cookie_cache_state {
...
+	return r;
+}
+INLINE void cache_entry_free(void *p) {
+	cache_entry *s = p;
+	if (!s)
+		return;
+	free(s->reply);
+	free(s->command);
+	free(s->callid);
+}
```
rfuchs, line 38: I guess you're missing the `free` for `s` itself

lbalaceanu:
Hi,

I also pushed a commit solving the observation that some code could be moved after a switch. Will test a bit more tomorrow, but more or less this would be my proposal for the code change.

Thank you

rfuchs:
LGTM now, thanks

lbalaceanu:
Hi, I also have no more comments to the code. Thanks.

rfuchs:
Actually one last request, to please document the new options in the man page :laughing:

lbalaceanu:
Hi, sure, just updated the docs.

lbalaceanu:
Hi @rfuchs, 

In case there is something more to do about this PR, please tell me.

Thank you

rfuchs:
> Hi @rfuchs,
> 
> In case there is something more to do about this PR, please tell me.
> 
> Thank you

Just waiting for internal review but I think I'll just push it through...

PR #1804, opened 2024-03-05
Fix TLS-ID change detection

Den4t:
Hi !

This fixes DTLS restarts, wrong condition are used, the same tls-id is detected as changed.
This error present in 11.x and 12.x versions.

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -2123,7 +2123,7 @@ static void __dtls_logic(const sdp_ng_flags *flags,
 		__fingerprint_changed(other_media);
 		other_media->fingerprint = sp->fingerprint;
 	}
-	else if (other_media->tls_id.len && (sp->tls_id.len || str_cmp_str(&other_media->tls_id, &sp->tls_id))) {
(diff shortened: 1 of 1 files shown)


daemon/call.c
```diff
@@ -2123,7 +2123,7 @@ static void __dtls_logic(const sdp_ng_flags *flags,
 		__fingerprint_changed(other_media);
 		other_media->fingerprint = sp->fingerprint;
 	}
-	else if (other_media->tls_id.len && (sp->tls_id.len || str_cmp_str(&other_media->tls_id, &sp->tls_id))) {
```
rfuchs, line 2126: I think this should actually be `(other_media->tls_id.len && (sp->tls_id.len == 0 || str_cmp_str(&other_media->tls_id, &sp->tls_id)))`

Den4t: I will test this in my environment.

Den4t: Hi !

I have tested changes, no errors, so i push it to this PR.

PR #1809, opened 2024-03-13
Some DTMF injection fixes

tombriden:
A few scenarios are fixed for me with these changes.

> packet_encoded_tx: add a ts delay when transmitting DTMF event packets

This fixes not passing a `ts_delay` into `codec_output_rtp` resulting in packets not being scheduled correctly (like already happens in `codec_add_dtmf_packet`) and an, eg, 100ms event has all packets transmitted in as few as 20ms

>  dtmf_event_payload: canonicalise DTMF end event ts if start packet send was delayed

in some scenarios DTMF injected with, eg, 100ms duration are transmitted much shorter due to the start time being adjusted but not the end time

> dtmf_inject: fix generating one too many event packets

the starting value used to add num_samples to is that of the current packet, which is already increased from the previous packet. The result is a 100ms injection coming through as 6 packets with a total event-duration of 960
dtmf_inject: adjust start_pts if last_event + pause is later than it

> codec_last_dtmf_event: return ts of dtmf_state if handler queue is empty
  dtmf_inject: adjust start_pts if last_event + pause is later than it

DTMF requires an inter-digit pause, which is allowed for with the `pause` parameter. However, this is only accounted for if an injection request comes in during an insertion, as its the event queue used to return the last event end timestamp. If there's nothing on the queue, we can fall back to the `dtmf_state` var as that seems to always be the most recent. 
The second commit can now use this value to calculate a minimum start ts for the injected event to ensure the pause is accounted for. Similar to the previous change, the actual time is increased by one packet's worth of samples to avoid an 100ms pause only having 80ms worth of packets

> dtmf: only update recv list if not injected and send list if injected, delayed or not blocked

when the DTMF-security mode is one of the values between `PCM_REPLACE_START/END`, the event packets are transcoded to PCM DTMF. Injected DTMF gets added to the `dtmf_recv` list and can cause out of order timestamp in the list, resulting in `is_in_dtmf` incorrectly returning `NULL`, resulting in those PCM DTMF frames leaking through.
I'm not sure if its needed, but to me it seems logical that maintaining separate lists for `recv` and `send` shouldn't result in them being basically identical by always adding all events to it (other than an adjusted ts for dtmf-delay). So have included a change only add to the send list if its injected, unblocked or being delayed. This way each list contains only contains the relevant events.

### DIFF
--- a/daemon/codec.c
+++ b/daemon/codec.c
@@ -2295,10 +2295,12 @@ static tc_code packet_dtmf(struct codec_ssrc_handler *ch, struct codec_ssrc_hand
 					ret = TCC_CONSUMED;
 				else
 					ret = packet_dtmf_fwd(ch, input_ch, dup, mp);
+				mp->ssrc_out->parent->seq_diff++;
 
 				if (ret != TCC_CONSUMED)
 					__transcode_packet_free(dup);
 			}
+			mp->ssrc_out->parent->seq_diff--;
 
 			// discard the received event
 			do_blocking = true;
--- a/daemon/codec.c
+++ b/daemon/codec.c
@@ -2666,7 +2666,7 @@ void codec_add_dtmf_event(struct codec_ssrc_handler *ch, int code, int level, ui
 uint64_t codec_last_dtmf_event(struct codec_ssrc_handler *ch) {
 	struct dtmf_event *ev = t_queue_peek_tail(&ch->dtmf_events);
 	if (!ev)
-		return 0;
+		ev = &ch->dtmf_state;
 	return ev->ts;
 }
 
--- a/daemon/codec.c
+++ b/daemon/codec.c
@@ -3979,6 +3979,7 @@ static void packet_encoded_tx(AVPacket *pkt, struct codec_ssrc_handler *ch, stru
 	// check special payloads
 
 	unsigned int repeats = 0;
+	unsigned long ts_delay = 0;
 	int payload_type = -1;
 	int dtmf_pt = ch->handler->dtmf_payload_type;
 	if (dtmf_pt == -1)
@@ -3994,6 +3995,12 @@ static void packet_encoded_tx(AVPacket *pkt, struct codec_ssrc_handler *ch, stru
 			ch->rtp_mark = 1; // DTMF start event
 		else if (is_dtmf == 3)
 			repeats = 2; // DTMF end event
+		// we need to pass a ts_delay to codec_output_rtp to ensure the calculated time
+		// to send the packet is offset by the event duration of the DTMF packets
+		// but we need to reduce it by one packet duration so that the delay is offset
+		// from the first event packet
+		struct telephone_event_payload *ev_pt = (void *) inout->s;
+		ts_delay = ntohs(ev_pt->duration) - (ch->handler->dest_pt.ptime * ch->handler->dest_pt.clock_rate / 1000);
 	}
 	else {
 		if (is_silence_event(inout, &ch->silence_events, pkt->pts, pkt->duration))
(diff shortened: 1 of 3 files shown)


rfuchs:
Looks like this is breaking some of the test cases. I'll try to adapt.

daemon/dtmf.c
```diff
@@ -853,11 +853,14 @@ const char *dtmf_inject(struct call_media *media, int code, int volume, int dura
...
-		start_pts = last_end_pts + pause * ch->dest_pt.clock_rate / 1000;
-	}
+	// get the last event end time, and increase by the required pause
+	// conversely to the above, we need to add the last packet num samples to its TS before adding
+	// a pause so we dont generate one packet too few
+	// if that's later than start_pts, we need to adjust it
+	uint64_t last_end_pts = codec_last_dtmf_event(csh) + ((pause + ch->dest_pt.ptime) * ch->dest_pt.clock_rate / 1000);
+	if (last_end_pts > start_pts)
+		start_pts = last_end_pts;
+
```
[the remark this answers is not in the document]
tombriden: ah, good spot. have fixed that to only adjust last_end_pts if it isn't 0

rfuchs:
I've pushed the final rebased series of commits (plus an extra one that was needed) including the updated tests to https://github.com/sipwise/rtpengine/tree/rfuchs/gh1809

Have a look if these look OK to you, then I will merge.

tombriden:
> I've pushed the final rebased series of commits (plus an extra one that was needed) including the updated tests to https://github.com/sipwise/rtpengine/tree/rfuchs/gh1809
> 
> Have a look if these look OK to you, then I will merge.

I don't have a scenario that makes use of the extra commit you've added, but it looks sane to me. Tests are passing here now too. Thanks a lot!

daemon/codec.c
```diff
@@ -3992,8 +3993,16 @@ static void packet_encoded_tx(AVPacket *pkt, struct codec_ssrc_handler *ch, stru
...
+		else {
+			// we need to pass a ts_delay to codec_output_rtp to ensure the calculated time
+			// to send the packet is offset by the event duration of the DTMF packets
+			// but we need to reduce it by one packet duration so that the delay is offset
+			// from the first event packet
+			struct telephone_event_payload *ev_pt = (void *) inout->s;
+			ts_delay = ntohs(ev_pt->duration) - (ch->handler->dest_pt.ptime * ch->handler->dest_pt.clock_rate / 1000);
+			if (is_dtmf == 3)
+				repeats = 2; // DTMF end event
+		}
```
rfuchs, line 4005: Do you think it would make sense to always do this TS adjustment, and not just for non-start event packets? I assume in your case the delay probably comes out as zero for the initial packet, but it's possible the first packet (triggering the start event) uses a different duration.

tombriden: I don't think so. the first event packet has had its ts increased from the previous packet already so it gets scheduled correctly; in the pcaps I made during testing, the first packet always looked right (~20ms from the previous non-event packet). Then, because the `ts_delay` is applied on top of that first event packet TS, this adjustment has to be one ptime less than the current event duration so that they don't end up being scheduled starting 40ms after the first event.

rfuchs: > I don't think so. the first event packet has had its ts increased from the previous packet already so it gets scheduled correctly; in the pcaps I made during testing, the first packet always looked right (~20ms from the previous non-event packet). Then, because the `ts_delay` is applied on top of that first event packet TS, this adjustment has to be one ptime less than the current event duration so that they don't end up being scheduled starting 40ms after the first event.

AFAICT the first event packet _usually_ has a duration of 20 ms or 160 samples, so that would come out as a delay of 160 - 20 * 8 = 0. But what if that packet got lost, and the packet triggering the start event (i.e. the state change) is actually the second one, using a delay of 40 ms or 320 samples?

tombriden: i gone ahead and done that anyway :)

rfuchs:
Everything merged and backported, thanks!

PR #1852, opened 2024-08-22
Added decoder mix "slots" concept, so that the desired wav channel ca…

petekelly:
…n be controlled when producing a mixed audio file

When a mixed wav file is created, the channels in the wav container are currently allocated in the same order as each SSRC is received, meaning it is impossible to know which channels have been allocated to the offer or answer side of the call. Furthermore if there is a reinvite or media file played, these are also allocated in the order that SSRC is received - so an "answer" could end up sharing a channel with an "offer" with no way of knowing this.

This patch allows you to specify how many channel slots should be allocated within the mixer, and allows you to then specify which slot is assigned to each media in the call (this will usually be 2 slots in total, slot 1 for answer, slot 2 for offer or vice versa).  

The mixer will then use up the slots as per these rules. So (assuming a MIX_MAX_INPUTS=4 limit in mix.h) channel 0,2 will be used for the offer and channel 1,3 will be used for the answer - subsequent SSRCs will rotate round and reuse 0,2 and 1,3 respectively.

If 3 slots are specified then slot 1 would use channels 0,3,6... Slot 2 would use channels 1,4,7... Slot 3 would use channels 2,5,8 assuming the MIX_MAX_INPUTS limits in mix.h allowed this.

If 0 or 1 slots are specified then channels will be used up as each SSRC is seen as per previous behaviour.

Note: I was unsure how to make/check defaults in call_interfaces.c so any pointers here would be appreciated. Also, all flags are set in the offer as it seems we can only guarantee this is available when the recording metafile is written - the answer may not yet exist.

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -2571,6 +2571,9 @@ static void __call_monologue_init_from_flags(struct call_monologue *ml, sdp_ng_f
 
 	call->last_signal = rtpe_now.tv_sec;
 	call->deleted = 0;
+	call->media_rec_slots = (flags->media_rec_slots > 0 && call->media_rec_slots == 0)
+		                        ? flags->media_rec_slots
+		                        : call->media_rec_slots;
 
 	// consume session attributes
 	t_queue_clear_full(&ml->sdp_attributes, sdp_attr_free);
@@ -2789,10 +2792,9 @@ static void __media_init_from_flags(struct call_media *other_media, struct call_
 	}
 
 	if (flags->opmode == OP_OFFER) {
+		ilog(LOG_DEBUG, "setting other slot to %u, setting slot to %u", flags->media_rec_slot_offer, flags->media_rec_slot_answer);
 		other_media->media_rec_slot = flags->media_rec_slot_offer;
 		media->media_rec_slot = flags->media_rec_slot_answer;
-		media->media_rec_slots = flags->media_rec_slots;
-		other_media->media_rec_slots = flags->media_rec_slots;
 	}
 }
 
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -2787,6 +2787,13 @@ static void __media_init_from_flags(struct call_media *other_media, struct call_
 		if (sp->desired_family)
 			media->desired_family = sp->desired_family;
 	}
+
+	if (flags->opmode == OP_OFFER) {
+		other_media->media_rec_slot = flags->media_rec_slot_offer;
+		media->media_rec_slot = flags->media_rec_slot_answer;
+		media->media_rec_slots = flags->media_rec_slots;
+		other_media->media_rec_slots = flags->media_rec_slots;
+	}
 }
 
 unsigned int proto_num_ports(unsigned int sp_ports, struct call_media *media, sdp_ng_flags *flags,
--- a/daemon/call_interfaces.c
+++ b/daemon/call_interfaces.c
@@ -1738,6 +1738,17 @@ static void call_ng_main_flags(sdp_ng_flags *out, str *key, bencode_item_t *valu
 		case CSH_LOOKUP("recording file"):
 			out->recording_file = s;
 			break;
+		case CSH_LOOKUP("recording-media-slot-offer"):
+			// This needs to be > 0
+			out->media_rec_slot_offer = bencode_get_integer_str(value, out->media_rec_slot_offer);
+			break;
+		case CSH_LOOKUP("recording-media-slot-answer"):
+			// This needs to be > 0
+			out->media_rec_slot_answer = bencode_get_integer_str(value, out->media_rec_slot_answer);
+			break;
+		case CSH_LOOKUP("recording-media-slots"):
+			out->media_rec_slots = bencode_get_integer_str(value, out->media_rec_slots);
+			break;
 		case CSH_LOOKUP("passthrough"):
 		case CSH_LOOKUP("passthru"):
 			switch (__csh_lookup(&s)) {
--- a/recording-daemon/mix.c
+++ b/recording-daemon/mix.c
@@ -31,7 +31,8 @@ struct mix_s {
 	CH_LAYOUT_T channel_layout[MIX_MAX_INPUTS];
 	AVFilterContext *amix_ctx;
 	AVFilterContext *sink_ctx;
-	unsigned int input_idx;
+	unsigned int next_idx[MIX_MAX_INPUTS]; //slots can never exceed MIN_MAX_INPUTS by definition
+	unsigned int channel_slots;
 	AVFrame *sink_frame;
 
 	resample_t resample;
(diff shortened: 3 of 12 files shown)


daemon/call.c
```diff
@@ -2787,6 +2787,13 @@ static void __media_init_from_flags(struct call_media *other_media, struct call_
 		if (sp->desired_family)
 			media->desired_family = sp->desired_family;
 	}
+
+	if (flags->opmode == OP_OFFER) {
+		other_media->media_rec_slot = flags->media_rec_slot_offer;
+		media->media_rec_slot = flags->media_rec_slot_answer;
+		media->media_rec_slots = flags->media_rec_slots;
```
rfuchs, line 2794: I assume the `media_rec_slots` value should be the same for all parties to a call within the same call? Perhaps store the value in the `call` instead then, and make sure it's set only once? It might make the logic in `mix_get_index` easier as well. Or you could even have it as a global config variable (in `rtpengine` or in `rtpengine-recording`)

rfuchs (changes requested):
A few comments. The slot selection logic in `mix_get_index` definitely is off and needs to be reworked.

daemon/call_interfaces.c
```diff
@@ -1738,6 +1738,17 @@ static void call_ng_main_flags(sdp_ng_flags *out, str *key, bencode_item_t *valu
...
 			out->recording_file = s;
 			break;
+		case CSH_LOOKUP("recording-media-slot-offer"):
+			// This needs to be > 0
+			out->media_rec_slot_offer = bencode_get_integer_str(value, out->media_rec_slot_offer);
+			break;
+		case CSH_LOOKUP("recording-media-slot-answer"):
+			// This needs to be > 0
+			out->media_rec_slot_answer = bencode_get_integer_str(value, out->media_rec_slot_answer);
+			break;
```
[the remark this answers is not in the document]
petekelly: So if the second argument is set to 1 then the default value (if used) would always come out at 1. Then there would need to be some check (perhaps directly within the case statement?) to check/set that the value was at least 1 (and output an error/warning log if the check happens)? Or is there a more elegant/standard place to do this?

daemon/call.c
petekelly: I wasn't keen on the global for `media_rec_slots`, in case the rtpengine instance was having to deal with multiple call scenarios (SIPREC with > 2 media, regular call with offer/answer etc). 

Yes it would be set to the same for all parties to a call within the same call. I'll look at storing it in the `call`. Could you point me to a sensible place to do this?

daemon/call_interfaces.c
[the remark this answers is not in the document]
rfuchs: > So if the second argument is set to 1 then the default value (if used) would always come out at 1. Then there would need to be some check (perhaps directly within the case statement?) to check/set that the value was at least 1 (and output an error/warning log if the check happens)? Or is there a more elegant/standard place to do this?

Using the "previous value" as second argument is fine. Using an explicit `1` as default value doesn't actually solve the problem, as it would still leave the value as zero if either the key/value isn't given at all, or if an explicit zero was given as value. So the code needs to explicitly check this. The most appropriate place to check if the value is sane is the place where it's being read or used, for example in `setup_stream_proc`

daemon/call.c
rfuchs: > I wasn't keen on the global for `media_rec_slots`, in case the rtpengine instance was having to deal with multiple call scenarios (SIPREC with > 2 media, regular call with offer/answer etc).
> 
> Yes it would be set to the same for all parties to a call within the same call. I'll look at storing it in the `call`. Could you point me to a sensible place to do this?

You can just put the values into `struct call` and leave the assignment in the same place. There isn't (yet) a dedicated function to assign `call` values from `flags`, but you could (ab)use `__call_monologue_init_from_flags()` and just access it through `ml->call`. There are some other call-specific flags assigned there already.

petekelly: @rfuchs  I think I am almost here with this. I am struggling to find/assign the "slot" from the offer/answer whenever a play_media happens. I can see that `call_play_media_ng` is called and that `setup_stream_proc` then gets called again to create some extra streams in the recording metafile. At this point I need to find the "slot" that is already set from the corresponding `media` or `monologue` but I can't see where this has to happen. Do you have any suggestions?

petekelly: @rfuchs please ignore the above, it was a bug in an OpenSIPS configuration when calling rtpengine...

recording-daemon/mix.c
```diff
@@ -82,32 +90,46 @@ static void mix_input_reset(mix_t *mix, unsigned int idx) {
 	mix->in_pts[idx] = 0;
 }
 
+unsigned int mix_get_index(mix_t *mix, void *ptr, unsigned int stream_channel_slot) {
+
+	ilog(LOG_DEBUG, "next index is set to %u for slot %u. channel slots for this mix are %u", mix->next_idx[stream_channel_slot], stream_channel_slot, mix->channel_slots);
+
+	if(mix->next_idx[stream_channel_slot] == mix_num_inputs+1) {
```
rfuchs, line 97: I think validation that `stream_channel_slot` is actually `< MIX_MAX_INPUTS` is missing? AFAICT there's nothing preventing somebody from setting `recording-media-slots=500` or something.

recording-daemon/mix.c
```diff
@@ -225,6 +247,12 @@ mix_t *mix_new(void) {
 	for (unsigned int i = 0; i < mix_num_inputs; i++)
 		mix->pts_offs[i] = (uint64_t) -1LL;
 
+	for (unsigned int i = 0; i < mix_num_inputs; i++) {
+		// initialise with the first mixer channel to use for each slot. This is set to mix_num_inputs+1
+		// so that we can detect first use and also if the maximum use has been reached.
+		mix->next_idx[i] = mix_num_inputs+1;
```
rfuchs, line 253: I think this could be made simpler by just initialising each `next_idx[i]` to `i` and then in `mix_get_index()` just get the current value of `next_idx[stream_channel_slot]`, use it as `next`, and then unconditionally increasing it by `mix->channel_slots`. Unless I'm reading the code wrong :grin:

rfuchs (commented):
Ok, I like this approach. Couple of minor concerns, otherwise this looks good

recording-daemon/mix.c
petekelly: Perfect suggestion, this is was I was trying to achieve originally -  but I got lost in over engineering it!

rfuchs:
I assume this replaces the `media_sdp_id` stuff present in current master (but not in 12.3)...

rfuchs (approved):

rfuchs:
> > I assume this replaces the `media_sdp_id` stuff present in current master (but not in 12.3)...
> > Edit: Actually I don't think it does 🤔
> 
> I agree I don't think it does - I think they are different conceptually. The stuff currently in master is switched on or off via the config file and allows SDP with multiple media lines to have their mix index set sequentially. (I assume for SIPREC with multiple streams). The patch I have put together lets you control per dialogue which "slot" each mix index should have.
> 
> I think they could/should probably co-exist. I wasn't sure of the path from main > LTS release so I just went down the route of developing against current LTS which is what we are using presently, however I can also apply my changes against current main/LTS if it is the correct procedure.

Yes, if you could try to reconcile the patches against master, then that would be much appreciated. This is a new feature and can't be backported against existing releases, so it can only go into master and future releases.

petekelly:
Done: https://github.com/sipwise/rtpengine/pull/1857

PR #1857, opened 2024-09-06
Added decoder mix "slots" concept, so that the desired wav channel ca…

petekelly:
…n be controlled when producing a mixed audio file

When a mixed wav file is created, the channels in the wav container are currently allocated in the same order as each SSRC is received, meaning it is impossible to know which channels have been allocated to the offer or answer side of the call. Furthermore if there is a reinvite or media file played, these are also allocated in the order that SSRC is received - so an "answer" could end up sharing a channel with an "offer" with no way of knowing this.

This patch allows you to specify how many channel slots should be allocated within the mixer, and allows you to then specify which slot is assigned to each media in the call (this will usually be 2 slots in total, slot 1 for answer, slot 2 for offer or vice versa).

Ported from https://github.com/sipwise/rtpengine/pull/1852

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -2630,6 +2630,9 @@ static void __call_monologue_init_from_flags(struct call_monologue *ml, struct c
 
 	call->last_signal = rtpe_now.tv_sec;
 	call->deleted = 0;
+	call->media_rec_slots = (flags->media_rec_slots > 0 && call->media_rec_slots == 0)
+								? flags->media_rec_slots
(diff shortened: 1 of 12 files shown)


a72:
@petekelly Could you please advise which parameters I should use in order to produce a WAV file with exactly two channels, where the offer is always written to channel 0 and the answer to channel 1?

Additionally, I would like this mapping to remain stable even if the SSRC changes during re-INVITEs, so that audio continues to be written to the same respective channels.

PR #1860, opened 2024-09-12
Honour the hosts IPv4 TTL and IPv6 hop limit

schlagert:
This patch forces rtpengine to use the IPv4 TTL and IPv6 hop limits set on the host for proxied packets in kernel mode.

Beforehand, the TTL was always set to a hardcoded value of 64. While this is the default almost everywhere, it may be desirable to be able to set a higher value. Especially when interfacing with complex carrier backbone networks.

### DIFF
--- a/kernel-module/xt_RTPENGINE.c
+++ b/kernel-module/xt_RTPENGINE.c
@@ -4988,6 +4988,9 @@ static int send_proxy_packet4(struct sk_buff *skb, struct re_address *src, struc
 	if (!net)
 		goto drop;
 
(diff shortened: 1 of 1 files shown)


schlagert:
Hi @rfuchs,

I just found out that the (IPv4) `sysctl_ip_default_ttl` was not available/exposed prior to kernel 4.6. If you need to support kernels older than that you'll need to `ifdef` accordingly. The IPv6 config seems to be save back to kernel 3.7 as far as I can see.

Sorry for the inconvenience.

PR #1870, opened 2024-10-18
Ensure (AEAD) req and copy pointers don't leak memory

sigxcpu76:
In `srtp_decrypt_aes_gcm` there are multiple places than can leak memory.

`req` pointer can leak when `IS_ERR` is true.
`copy` pointer is allocated on each iteration loop but it is free-ed only once outside the loop. If the loop runs more than once and it performs an early exit on `req` checks or it simply runs multiple times, the `copy` pointer will leak.

### DIFF
--- a/kernel-module/xt_RTPENGINE.c
+++ b/kernel-module/xt_RTPENGINE.c
@@ -5755,6 +5755,7 @@ static int srtp_decrypt_aes_gcm(struct re_crypto_context *c,
 		// make copy of payload in case the decyption clobbers it
 		if (!copy)
 			copy = kmalloc(r->payload_len, GFP_ATOMIC);
+			
 		if (copy)
 			memcpy(copy, r->payload, r->payload_len);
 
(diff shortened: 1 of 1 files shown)


sigxcpu76:
I am on the fence at likne on `aead_request_free(req);` because I see it passed down below to `PTR_ERR`. I cannot find that macro in the rtpengine source code, so I don't know what it does. If it is referenced somehow, this will be a "use after free". If it is not freed and `PTR_ERR` doesn't do it either then it will be a leak (maybe low probability but it is there).

rfuchs:
> I am on the fence at line 5752 on `aead_request_free(req);` because I see it passed down below to `PTR_ERR`. I cannot find that macro in the rtpengine source code, so I don't know what it does. If it is referenced somehow, this will be a "use after free". If it is not freed and `PTR_ERR` doesn't do it either then it will be a leak (maybe low probability but it is there).

`PTR_ERR` and `IS_ERR` are from the kernel itself. When `IS_ERR(req)` is true, then `req` is not something that was actually allocated, but rather just an error code. So you shouldn't try to free `req` in that case.

PR #1873, opened 2024-10-21
clamp the intercept stream max_packets between 0 and the value specif…

sigxcpu76:
It looks like the `info.max_packets` is corrupted here and != 0. This means that it is not capped to the `stream_packets_list_limit` value. I've found that its value (not sure if on each restart) is always `32647` which leads to a giant memory usage with many recorded calls.

This patch keeps the `==0` check and adds another upper boundary check to make sure we cap it to the max value specified in the `stream_packets_list_limit` kernel parameter (10 by default).


rfuchs:
Nice find. But I would say the more appropriate fix is to make sure the value is always initialised correctly. This should be done in `kernel_add_intercept_stream()`

PR #1880, opened 2024-11-29
Fix EL build errors

asolovjov:
* Added libatomic to the el/ build dependencies. This was required since commit 741f6ac when -latomic was added to lib/common.Makefile
* Changed link to the copr repo with spandsp3. There is seperate repo for EL9 now.


rfuchs:
Thank you

PR #1884, opened 2024-12-12
DTLS groups: align with OpenSSL 1.1.1 and common EC groups

space88man:
Two tweaks in DTLS configuration of supported groups:
- the 0x30000000L code path applies to 0x10101000L (OpenSSL 1.1.1) as well
- allow more EC groups, to conform to a very common set of OpenSSL defaults (P-521:P-384:P-256); ATM the only supported group configured is P-256

Background:
In the wild encountered some custom DTLS peers that wanted P-384 / P-521


space88man:
Closed with commit 27ea6b13.

Thank you.

PR #1889, opened 2024-12-19
Prevent warning logs when recording-method is pcap

orgads:
If recording-method is pcap, then proc is zero-initialized, so meta_filepath is empty. This shows many logs such as:

```
[core] Failed to open recording metadata file '(null)' for writing: Bad address
```

Prevent them by returning earlier.


orgads:
Still happens. Need to add protection also in `vappend_meta_chunk`.

PR #1899, opened 2025-01-15
Added several useful fields for mqtt stats publishing

Den4t:
Hi !

This PR adds several additional but useful fields to mqtt statistics:

- stream endpoint addtress
- stream endpoint port
- crypto suite (if used)
- transcoding status


Den4t:
Hi !

Thanks for fast merge. It is possible to do backport to 12.5 ?

rfuchs:
I don't consider this a bug fix

PR #1907, opened 2025-01-24
Doc: Fix misspelled reject-ICE offer attribute

orgads:


rfuchs:
Both are actually accepted by the parser. In fact any space is translated to a hyphen when parsing.

orgads:
Then why are there duplicate cases for other flags? A few examples:

```c++
  case CSH_LOOKUP("address family"):
  case CSH_LOOKUP("address-family"):
  // ...
  case CSH_LOOKUP("call-id"):
  case CSH_LOOKUP("call-ID"):
  case CSH_LOOKUP("call id"):
  case CSH_LOOKUP("call ID"):
  // ...
```

rfuchs:
> Then why are there duplicate cases for other flags? A few examples:

These are in the main parsing entry point, which doesn't have a call to `str_hyphenate`. Should probably streamline these and avoid `str_hyphenate` where possible and instead list non-hyphenated alternatives explicitly.

orgads:
I see, thanks for the clarification.

PR #1912, opened 2025-03-03
Force transcoding flag

hajekmi:
I have extended the daemon with flag force-transcoding to ensure that it will always transcode. This ensures that the loss rate between A-leg and B-leg is measured. If this flag comes into the NG protocol, it will always transcode (for example between G722 and G722 codec).


rfuchs (commented):
Please mention this flag in the docs `docs/ng_control_protocol.md`

hajekmi:
I commited patch

PR #1924, opened 2025-04-06
Prevent warning logs when recording-method is pcap #2

orgads:
If recording-method is pcap, then proc is zero-initialized, so meta_filepath is empty. This shows many logs such as:

[core] Failed to open recording metadata file '(null)' for writing: Bad address

Prevent them by returning earlier.

Fixes #1889 for real this time.

Amends commit 759fd72dc6866b2c89eddd8940bb4b2040699f23.

### DIFF
--- a/daemon/recording.c
+++ b/daemon/recording.c
@@ -835,6 +835,9 @@ static int vappend_meta_chunk(struct recording *recording, const char *buf, unsi
 		const char *label_fmt, va_list ap)
 {
 	struct iovec iov;
+	if (!recording->proc.meta_filepath)
+		return -1;
+
 	iov.iov_base = (void *) buf;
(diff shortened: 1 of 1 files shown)


orgads:
The error is unrelated to my changes.

orgads:
> Why not just put the check in open_proc_meta_file and do the early return there?

To avoid needless work until we reach there. Do you consider it negligible, and prefer to move it there?

orgads:
@rfuchs ping

I double-checked, and there are no other calls, so this should seal this case without extra work. If you still prefer to move the condition to `open_proc_meta_file` (and remove the previous fix as well), let me know and I'll update the PR.

rfuchs:
Sorry for the delay, completely lost focus on this

PR #1938, opened 2025-05-09
some fixes when using delay-buffer + block-media + play-media

tombriden:
when media playback happens, all packets get scheduled at the same time with a send time in the future. however, this results in the output_skew calculation triggering an unnecessary shift in the send timer, which can cause choppy playback

### DIFF
--- a/daemon/codec.c
+++ b/daemon/codec.c
@@ -1466,7 +1466,7 @@ void __codec_handlers_update(struct call_media *receiver, struct call_media *sin
 	if (proto_is_not_rtp(receiver->protocol)) {
 		__generator_stop_all(receiver);
 		__generator_stop_all(sink);
-		codec_handlers_stop(&receiver->codec_handlers_store, sink);
+		codec_handlers_stop(&receiver->codec_handlers_store, sink, a.clear_delay_buffer);
 		return;
 	}
 
@@ -1482,7 +1482,10 @@ void __codec_handlers_update(struct call_media *receiver, struct call_media *sin
 	// we're doing some kind of media passthrough - shut down local generators
 	__generator_stop(receiver);
 	__generator_stop(sink);
-	codec_handlers_stop(&receiver->codec_handlers_store, sink);
+
+	// clear_delay_buffer is enabled by play-media to ensure media packets don't get scheduled
+	// alongside the delay buffer packets resulting in choppy audio
+	codec_handlers_stop(&receiver->codec_handlers_store, sink, a.clear_delay_buffer);
 	// XXX this can cause unnecessary shutdown/resets of codec handlers
 
 	bool is_transcoding = false;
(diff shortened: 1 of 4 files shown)


daemon/codec.c
```diff
@@ -2485,7 +2488,11 @@ void codec_output_rtp(struct media_packet *mp, struct codec_scheduler *csch,
 
 	ts_diff_us = p->ttq_entry.when - rtpe_now;
 
-	csch->output_skew = csch->output_skew * 15 / 16 + ts_diff_us / 16;
+	if (csch->is_media_playback) {
+		ilogs(transcoding, LOG_DEBUG, "Skipping clock skew calculation during media playback");
```
rfuchs, line 2492: This would be printed for every single packet if debugging is enabled. Either add the `LOG_FLAG_LIMIT` modifier or just remove it (maybe just log it once when the flag is set)

tombriden: good shout, was handy while i was testing but not needed now. have added a line when the flag is set instead

tombriden:
I've added a couple more commits to fix some other issues we were having

* fix not processing DTMF events while media is being blocked and DTMF-security=silence

as per the commit message, incoming DTMF would not be processed and sent over the `dtmf-log-dest` / `dtmf-log-ng-tcp` if the `dtmf_block_mode` was silence and media was being blocked. By updating `packet_dtmf` and `handler_dtmf_func` we can skip `handler_func_transcode` for DTMF replacement packets while media is blocked, and ensure they're still not transmitted

* add flag to discard delay buffer packets in codec_handlers_stop before media playback
when we use the `play-media` functionality, we first have to `block-media` to prevent both SSRCs being transmitted causing really choppy audio. Even if `delay-buffer` is used, the media playback packets are scheduled immediately so the delay-buffer packets end up being sent at the same time, causing some choppiness in the audio. I've added a flag that's always set to true on media-playback that ensures the delay-buffer queue is cleared, rather than flushed when the codec handlers are updated before playback starts. 

There are still a few silence packets transmitted, but its much cleaner than it was. I'm guessing these one's are packets added to the scheduler timer queue? but not sure how to clear those.


Would appreciate some feedback if you think this would be better solved a different way, otherwise can these commits please be applied to the mr11.5 branch if accepted?

Thanks

include/call_interfaces.h
```diff
@@ -167,6 +167,7 @@ struct sdp_ng_flags {
 	long long duration;
 	long long pause;
 	long long start_pos;
+	bool flush_delay_buffer;
```
rfuchs, line 170: The `flags` struct is usually reserved for things set by the user / controlling agent, so I'd prefer to pass this bool down some other way if at all possible.  How about adding it to `struct chu_args`, which would make it available in the `a` argument, and then pass it as an argument to `codec_update_all_source_handlers`?

tombriden: ok, yeah that makes sense, thanks.
Have made the change but flipped the bool to be `clear_delay_buffer` as it will be false when not explicitly passed in

rfuchs:
Rebased and merged, thanks

PR #1979, opened 2025-07-28
add dtls handshake debug log

wangduanduan:
add dtls handshake debug log

### DIFF
--- a/daemon/dtls.c
+++ b/daemon/dtls.c
@@ -600,7 +600,7 @@ static int try_connect(struct dtls_connection *d) {
 			if (d->connected) {
 				ilogs(crypto, LOG_INFO, "DTLS data received after handshake, code: %i", code);
 			} else {
-				ilogs(crypto, LOG_DEBUG, "DTLS handshake not completed yet, current state: %s", 
+				ilogs(crypto, LOG_DEBUG, "DTLS handshake not completed yet, current state: %s",
 					SSL_state_string_long(d->ssl));
 			}
 			break;
--- a/include/dtls.h
+++ b/include/dtls.h
@@ -67,6 +67,7 @@ void dtls_shutdown(struct packet_stream *ps);
 
 
 
+
 INLINE void __dtls_hash(const struct dtls_hash_func *hash_func, X509 *cert, unsigned char *out,
 		unsigned int bufsize)
 {


daemon/dtls.c
```diff
@@ -582,6 +582,9 @@ static int try_connect(struct dtls_connection *d) {
 		case SSL_ERROR_WANT_WRITE:
 			if (d->connected) {
 				ilogs(crypto, LOG_INFO, "DTLS data received after handshake, code: %i", code);
+			} else {
+				ilogs(crypto, LOG_INFO, "DTLS handshake not completed yet, current state: %s",
```
rfuchs, line 586: This should be `DEBUG` level I would say. Otherwise it would be too noisy to log every time.

daemon/dtls.c
```diff
@@ -637,8 +640,18 @@ static long dtls_bio_callback(BIO *bio, int oper, const char *argp, size_t len,
 	const endpoint_t *fsin = &ps->endpoint;
 	if (fsin->port == 9 || fsin->address.family == NULL)
 		return ret;
+	
+	if(len > 13 && (unsigned char)argp[0] == DTLS_CT_HANDSHAKE) {
+		ilogs(srtp, LOG_DEBUG, "Sending DTLS handshak %02x %s packet to %s",
```
rfuchs, line 645: Typo in `handshake`

daemon/dtls.c
```diff
@@ -980,3 +1003,49 @@ void dtls_connection_cleanup(struct dtls_connection *c) {
 	}
 	ZERO(*c);
 }
+
+
+const char *dtls_content_type_str(unsigned char type) {
```
rfuchs, line 1008: These two functions are private to `dtls.c` and so should be `static`

include/dtls.h
```diff
@@ -15,6 +15,25 @@
 #define DTLS_MAX_DIGEST_LEN 64
 #define DTLS_MTU_OVERHEAD 48 // 40 bytes IPv6 + 8 bytes UDP
 
+#define DTLS_CT_CHANGE_CIPHER_SPEC 20
```
rfuchs, line 18: These defines could go into `dtls.c` as well as they're not used anywhere else

rfuchs (commented):
Good improvement, thanks, a few minor comments

PR #1982, opened 2025-07-30
add timestamp markers to recording metadata for pause/resume events

orgads:
Write human-readable timestamp markers to the metadata file whenever recording is paused or resumed via control protocol commands.

Include ISO-format timestamps with millisecond precision that aid in tracking changes in recording state.

Example metadata output:
  Recording paused at: 2025-07-29T15:30:45 (1234567.890 ms)
  Recording resumed at: 2025-07-29T15:31:20 (1234598.123 ms)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/daemon/recording.c
+++ b/daemon/recording.c
@@ -402,7 +402,22 @@ void recording_start_daemon(call_t *call) {
 }
 // lock must be held
 void recording_start(call_t *call) {
+	bool was_paused = call->recording && !CALL_ISSET(call, RECORDING_ON);
 	CALL_SET(call, RECORDING_ON);
+
+	// Write resume timestamp to metadata file if resuming a paused recording
+	if (was_paused && call->recording->pcap.meta_fp) {
+		time_t now = rtpe_now / 1000000;
+		char timebuffer[32];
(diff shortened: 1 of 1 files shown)


daemon/recording.c
```diff
@@ -402,7 +402,23 @@ void recording_start_daemon(call_t *call) {
...
+	if (was_paused && call->recording->pcap.meta_fp) {
+		time_t now = rtpe_now / 1000000;
+		char timebuffer[32];
+		struct tm timeinfo;
+		if (localtime_r(&now, &timeinfo) != NULL) {
+			strftime(timebuffer, sizeof(timebuffer), "%FT%T", &timeinfo);
+			fprintf(call->recording->pcap.meta_fp,
+				"\nRecording resumed at: %s (%.3lf ms)\n",
+				timebuffer, rtpe_now / 1000.0);
+			fflush(call->recording->pcap.meta_fp);
```
rfuchs, line 418: `fflush` isn't really needed since the file is unbuffered already, but can leave it if you prefer

PR #1993, opened 2025-08-20
Add audio raw RTP file playback capability

Fr-Soltanzadeh:
Hi,

This PR adds the capability to play audio directly from raw RTP payload file to improve performance .

New Command Parameters
The existing play_media NG command is extended with two new parameters:

audio-raw_rtp_file: Path to the file containing raw RTP payloads (no headers).

audio-raw_rtp_codec: Codec of the payloads (e.g., PCMU, PCMA).

Example Usage:

bash
curl -X POST http://localhost:2225/ng-plain -H "Content-Type: application/json" -d '{
    "command": "play media",
    "audio-raw-rtp-file": "/path/to/file.raw", 
    "audio-raw-rtp-codec":"PCMU", 
    "call-id": "123abc", 
    "from-tag": "456def"
}'

Codec Support: Initially supports G.711 PCMU and PCMA

All existing functionality remains unchanged. The implementation includes proper error handling and memory management.

### DIFF
--- a/daemon/media_player.c
+++ b/daemon/media_player.c
@@ -1104,57 +1104,6 @@ static int media_player_find_file_begin(struct media_player *mp) {
 static bool media_player_read_packet(struct media_player *mp) {
 	if (!mp->coder.fmtctx)
 		return true;
-	// Handle raw RTP file playback
-    if (mp->coder.audio_raw_rtp_mode) {
-        // Check if we have more data
-        if (mp->coder.audio_raw_rtp_pos >= mp->coder.audio_raw_rtp_data.len) {
-            ilog(LOG_DEBUG, "End of raw RTP file reached");
-            return true;
-        }
-        
-        // Calculate bytes to read (one frame)
-        size_t bytes_to_read = MIN(mp->coder.audio_raw_rtp_frame_size, 
-                                  mp->coder.audio_raw_rtp_data.len - mp->coder.audio_raw_rtp_pos);
-        // Allocate packet
-        mp->coder.pkt = av_packet_alloc();
-        if (!mp->coder.pkt) {
-            ilog(LOG_ERR, "Failed to allocate packet");
-            return true;
-        }
-        
-        // Fill packet with raw RTP data
-        if (av_new_packet(mp->coder.pkt, bytes_to_read) < 0) {
-            ilog(LOG_ERR, "Failed to create packet");
-            av_packet_free(&mp->coder.pkt);
-            return true;
-        }
-        if (bytes_to_read > 0) {
-        	memcpy(mp->coder.pkt->data, 
-				   mp->coder.audio_raw_rtp_data.s + mp->coder.audio_raw_rtp_pos,
-				   bytes_to_read);
-		}
-		if (bytes_to_read < mp->coder.audio_raw_rtp_frame_size) {
-			// Pad with silence if needed
-			memset(mp->coder.pkt->data + bytes_to_read, 
-				mp->coder.silence_byte,
-				mp->coder.audio_raw_rtp_frame_size - bytes_to_read);
-			ilog(LOG_DEBUG, "Padding %zu bytes of silence",
-				mp->coder.audio_raw_rtp_frame_size - bytes_to_read);
-		}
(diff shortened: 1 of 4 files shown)


daemon/media_player.c
```diff
@@ -1104,6 +1104,57 @@ static int media_player_find_file_begin(struct media_player *mp) {
 static bool media_player_read_packet(struct media_player *mp) {
 	if (!mp->coder.fmtctx)
 		return true;
+	// Handle raw RTP file playback
+    if (mp->coder.audio_raw_rtp_mode) {
```
rfuchs, line 1108: Since `media_player_read_packet` is already used as `run_func` function pointer, this whole block would lend itself to also be its own function and then be used as `run_func`, to avoid the if/else

daemon/media_player.c
```diff
@@ -1475,6 +1527,97 @@ static mp_cached_code __media_player_add_file(struct media_player *mp,
...
+    fclose(f);
+    
+    // Store in player context
+    mp->coder.audio_raw_rtp_data.s = file_data;
+    mp->coder.audio_raw_rtp_data.len = file_size;
+    mp->coder.audio_raw_rtp_pos = 0;
+    // Set codec parameters based on input
+    if (opts.audio_raw_rtp_codec.len == 4 && strncasecmp(opts.audio_raw_rtp_codec.s, "PCMU", 4) == 0) {
+        mp->coder.audio_raw_rtp_codec = AV_CODEC_ID_PCM_MULAW;
+        mp->coder.audio_raw_rtp_frame_size = 160; // 20ms frames
```
rfuchs, line 1569: Maybe make this dependent on the `ptime`?

daemon/media_player.c
```diff
@@ -1475,6 +1527,97 @@ static mp_cached_code __media_player_add_file(struct media_player *mp,
...
+        ilog(LOG_ERR, "Failed to open raw RTP file: %s", file_path);
+        return false;
+    }
+    
+    // Get file size
+    fseek(f, 0, SEEK_END);
+    long file_size = ftell(f);
+    fseek(f, 0, SEEK_SET);
+    // Read entire file into memory
+    char *file_data = malloc(file_size);
```
rfuchs, line 1553: We tend to use `g_malloc` and `g_free` in the code.

Also glib provides a convenient `g_file_get_contents()` helper to do just this.

daemon/media_player.c
```diff
@@ -1104,6 +1104,57 @@ static int media_player_find_file_begin(struct media_player *mp) {
...
+				mp->coder.silence_byte,
+				mp->coder.audio_raw_rtp_frame_size - bytes_to_read);
+			ilog(LOG_DEBUG, "Padding %zu bytes of silence",
+				mp->coder.audio_raw_rtp_frame_size - bytes_to_read);
+		}
+        mp->coder.audio_raw_rtp_pos += bytes_to_read;
+        
+        // Simulate packet timing (20ms per frame)
+		mp->coder.pkt->pts = mp->last_frame_ts;
+        mp->coder.pkt->duration = 160;
```
rfuchs, line 1148: This should probably depend on either `ptime` or the frame size variable?

rfuchs (commented):
Ok. A few comments for you to consider. But I can merge this as is if you prefer.

I do wonder though if this couldn't have been achieved through some other means, for example by telling ffmpeg (libav...) that it is in fact opening a raw file. It does support both raw µ and A law input files, and for codecs that require framing (Opus...) it will need some more sophisticated processing anyway.

daemon/media_player.c
```diff
@@ -1475,6 +1527,97 @@ static mp_cached_code __media_player_add_file(struct media_player *mp,
...
+    if (!file_data || fread(file_data, 1, file_size, f) != file_size) {
+        ilog(LOG_ERR, "Failed to read raw RTP file");
+        fclose(f);
+        free(file_data);
+        return false;
+    }
+    fclose(f);
+    
+    // Store in player context
+    mp->coder.audio_raw_rtp_data.s = file_data;
```
rfuchs, line 1563: Actually I think this is never freed? So that definitely needs to be fixed

rfuchs:
Reopen when needed

PR #1998, opened 2025-08-28
EL RPM fixes

ipoddubny:
1. Syntax error in the spec file, `if` instead of `%if`, rpmbuild fails.
2. Service unit with type=notify requires the service to be built with systemd support, fixed by adding a BuildRequires dependency.
3. DKMS module fails to build because of missing `gen-rtpengine-kmod-flags`, fixed by adding the file to the package.


rfuchs:
Thank you. I suppose this should be backported as well.

PR #2008, opened 2025-10-03
recording: only update metafile offset after successfully parsing the change

tombriden:
when under load it's possible that a metafile gets updated by rtpengine while it's still being read by the recording daemon. this can result in a WARNING being logged and the loop breaking. That section ends up unprocessed and any changes in it don't get applied. This change only updates the offset after a successful section read. Therefore, if a section is incomplete and the loop breaks, the next inotify event will retry that incomplete section and it should succeed

### DIFF
--- a/recording-daemon/metafile.c
+++ b/recording-daemon/metafile.c
@@ -395,8 +395,6 @@ void metafile_change(char *name) {
 		g_string_append_len(s, buf, ret);
 	}
 
-	// save read position and close file
-	mf->pos = lseek(fd, 0, SEEK_CUR);
 	close(fd);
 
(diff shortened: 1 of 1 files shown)


tombriden:
we're using `tls-send-to`, and enable forwarding of one channel on after `offer` and the other after `answer`. Under load, we've found this can result in only one channel being forwarded and `Content truncated in section...` being logged due to the metafile change that adds `STREAM 0 FORWARD 1` not being processed.

tombriden:
thinking about this, the 2 cases to check for an unexpected `NUL` char shouldn't be handled in the same way or it'll never finish processing the file

PR #2033, opened 2025-12-01
Add inbound-peer, outbound-peer and peer flags for automatic interface selection

xadhoom:
Opening this for discussion.

In loosely-coupled or heterogeneous/dynamic (cough kubernetes cough) multihomed environments is often hard to understand which interface to use. Ideally this should be doable if "in advance" is possibile to understand which rtpengine instance is going to be used, which is not easy nor handled by the protocol right now.

<strike>
I'm proposing to introduce a new direction flag format `auto-to-<ip>` that automatically selects the matching configured rtpengine interface based on the system's routing table. `<ip>` can be either IPv4 or IPv6.~~

~~When a direction flag contains `auto-to-<ip>`, the system:
- Uses a temporary UDP socket to determine the local address assigned by routing (like is done by freeswitch or janus or others)
- Finds the first configured rtpengine interface matching that local address
- Uses that interface for the media stream (effectively replacing the `auto-to-<ip>` given in the direction with the real configured interface name)
</strike>

I'm proposing to introduce a ~two~ three new NG protocol flags that automatically select the matching configured rtpengine interface based on the system's routing table:
  
    - inbound-peer=<ip>: resolves to interface for inbound (from) direction
    - outbound-peer=<ip>: resolves to interface for outbound (to) direction
    - peer=<ip>: resolves to interface when only one interface is required
   
When these flags are used, the system:
    - Uses a temporary UDP socket to determine the local address assigned by routing
    - Finds the first configured rtpengine interface matching that local address
    - Uses that interface for the media stream

This allows for dynamic interface selection based on network topology without requiring explicit interface names in the signaling.

<strike>
The change is fully backward compatible and will not add any additional processing if the interfaces keeps original names (unless they are named `auto-to-<something>`)

The usage is meant to provide the rtp destination ip address in `auto-to-<ip>` from the caller, and how is computed is out of this scope. (On answer it can be computed by peer sdp, but on offer depends on the underlying logic, so I've decided to keep this decision external).
</strike>

The change is fully backward compatible.

In our use case we have rtpengine nodes running in network mode host across a kubernetes cluster, which gets automatically configured with all available interfaces of the host, because that cluster is multihomed and may receive traffic on any side. Right now we've "solved" it by having a sidecar container within each rtpengine which does exactly the same, but this breaks the possibility to add multiple rtpengines to each kamailio node, because we don't know which node is going to be used before calling manage/offer/etc.

What do you think?

### DIFF
--- a/daemon/media_socket.c
+++ b/daemon/media_socket.c
@@ -45,7 +45,6 @@
 #define MAX_RECV_LOOP_STRIKES 5
 #endif
 
-
 TYPED_GQUEUE(logical_intf, struct logical_intf)
 
 struct intf_key {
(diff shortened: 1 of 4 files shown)


daemon/media_socket.c
```diff
@@ -45,6 +45,9 @@
 #define MAX_RECV_LOOP_STRIKES 5
 #endif
 
+#define AUTO_DIRECTION_PREFIX "auto-to-"
+#define AUTO_DIRECTION_PREFIX_LEN (sizeof(AUTO_DIRECTION_PREFIX) - 1)
```
rfuchs: Have you considered using a dedicated dictionary key for this, instead of re-using the interface selection key with a magic prefix string? It's fine either way, but perhaps it would make for a cleaner interface.

So e.g. for interface selection we have `direction=` or alternatively `from-interface=` and `to-interface=`, or just `interface=` for other methods. And then to use address-based interface selection, you could have `from-address=` and `to-address=` or something similar. Just an idea, haven't looked at how much more trouble this would be.

rfuchs (commented):
This is a nice feature and in fact I've had something similar in mind already, although my idea would have been a list of routes statically configured in the config file (or alternatively, pulled from the OS at startup), and then the interfaces would be determined from this configured routing table. I assume in your use case the actual underlying routing tables can change dynamically, possibly at runtime?

A few comments for your consideration, but I could merge this as it is. Up to you.

Add to the documentation please. Some tests would be nice as well, but I'm not sure if the preload library providing the mock network environment is able to support this. It might need some extra work.

xadhoom:
Sorry I forgot to add the case for the single interface, if the approach is ok will do tomorrow.

PR #2056, opened 2026-01-23
CI: Build docker on PR and push when tags are created

orgads:


mika:
@orgads maybe you want to take a look at https://github.com/sipwise/rtpengine/pull/2063 and see whether that would be useful for you?

mika:
@orgads see e.g. https://github.com/sipwise/rtpengine/releases/tag/mr13.5.1.9 - is that useful for you or is there still any need for docker builds?

orgads:
@mika Thank you! This is very close to what I need. If I install `libavcodec-extra`, will the extra codecs like AMR be supported, or does it need to be available on build time?

And can you also publish for arm64?

orgads:
Ok, I found the answer for AMR - it works (if I install `libavcodec-extra`). So only arm64 is an open issue.

```
                 AMR: fully supported
              AMR-WB: fully supported
```

Thanks again!

orgads:
Pushed a PR for arm64:
- #2094

rfuchs:
Ok, closing this then

PR #2093, opened 2026-04-10
Build: Remove unused packages in docker build

orgads:


Dockerfile
```diff
@@ -61,9 +57,6 @@ EXPOSE 23000-65535/udp 22222/udp
 RUN apt-get update \
   && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
   curl \
-  iproute2 \
-  iptables \
-  libglib2.0-0 \
```
rfuchs, line 66: Sure about this? Because rtpengine definitely uses it

orgads: Which one? `libglib2.0-0`? It is a dependency of `libjson-glib-1.0-0` anyway, so it is installed. Do you use it directly too?

rfuchs: Ok, yes, rtpengine heavily relies on GLib, but fine if it's installed as dependency

orgads: Then I'll keep it for correctness (and in case `libjson-glib` will be dropped).

orgads: Done, and added also `libglib2.0-dev` for build stage.

PR #2094, opened 2026-04-10
CI: Build also arm64 debian packages

orgads:


.github/workflows/debian-package-build.yml
```diff
@@ -63,6 +65,14 @@ jobs:
...
             debian/artifacts/*.changes
             debian/artifacts/*.buildinfo
+      - name: Archive arm64 build result
+        if: matrix.os == 'ubuntu-24.04-arm'
+        uses: actions/upload-artifact@v6
+        with:
+          name: packages-${{ matrix.dist }}
+          if-no-files-found: error
+          retention-days: 14
+          path: debian/artifacts/*_arm64.deb
```
orgads, line 75: This is a bit ugly, but required to prevent duplicate file errors. Feel free to suggest an alternative.

orgads: Actually the duplicate name was for `name` prop. I renamed it for arm. So it might be possible to use a single step with `matrix.os` appended to `name`, but I'm not sure it's worth it. It will also store unneeded artifacts.

PR #2101, opened 2026-05-04
Fix deb build

orgads:
```
/usr/include/mosquitto/libcommon_cjson.h:31:10: fatal error: cjson/cJSON.h: No such file or directory
   31 | #include <cjson/cJSON.h>
      |          ^~~~~~~~~~~~~~~
```

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -15,6 +15,7 @@ Build-Depends:
  libavutil-dev (>= 6:10),
  libbcg729-dev <!pkg.ngcp-rtpengine.nobcg729>,
  libbencode-perl,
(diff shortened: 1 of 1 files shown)


rfuchs:
Which Debian version are you seeing this on? Looks like this should rather be fixed on the Debian side?

Edit: OK I see the failures, looks like Debian Sid broke their packaging.

rfuchs:
Reported here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1135697

orgads:
Thank you

PR #2105, opened 2026-05-11
support 'all' codec accept mode

orgads:
Add the 'all' keyword for the codec accept option in publish mode:
- 'all': keep all codecs supported for transcoding and their
  supplemental codecs, removing only unsupported ones
- 'all' + 'any': keep all codecs from the offer without removing any


orgads:
I think to simplify this to just `all`, and pass both `all` and `any` if we want to include also unsupported coders. What do you say?

rfuchs:
> I think to simplify this to just `all`, and pass both `all` and `any` if we want to include also unsupported coders. What do you say?

Good idea!

orgads:
Done

PR #2114, opened 2026-05-27
ignore rtcp-based packet loss in individual ssrc streams

jmontorosf:
At least in this version of RTPengine (13.5.1) there are 3 different metrics for packet loss that can be exported via MQTT:
* in voipmetrics, based on RTCP packets
* in rate, that requires kernel mode to work, but is not feasible for us at the moment
* for each ssrc, even in userspace mode can measure the detected packet loss locally

Currently the one that fits better the our goal, to measure packets lost per second in userspace it would be the last one and if we want to have a general view we can sum them up together so we can see all the lost packets on a particular interface...

However, there is a catch, that individual ssrcs are not only measuring local packet loss but also accounting for RTCP messages and the problem is when an RTCP packet, it resets the local counters

image

you can see the red line is just counting rtcp messages, because it's not actually detecting local loss, and the orange one is increasing steadily for local counters, but dropping whenever there is a RTCP packet received breaking the packet lost per second

I've come across this line in this PR, and disabling it, it works like a charm.

Is there any reason for that? Could we define different metrics for rtcp based packet loss and local measured packet loss, worst case scenario could there be a way to enable/disable rtcp metrics with a flag?

Thanks

### DIFF
--- a/daemon/mqtt.c
+++ b/daemon/mqtt.c
@@ -240,22 +240,24 @@ static void mqtt_ssrc_stats(struct ssrc_entry_call *ssrc, JsonBuilder *json, str
 	json_builder_begin_object(json);
 
 	// copy out values
-	int64_t packets, octets, packets_lost, duplicates;
+	int64_t packets, octets, packets_lost, packets_lost_rtcp, duplicates;
 	packets = atomic64_get_na(&ssrc->stats->packets);
 	octets = atomic64_get_na(&ssrc->stats->bytes);
 	packets_lost = ssrc->packets_lost;
+	packets_lost_rtcp = ssrc->packets_lost_rtcp;
 	duplicates = ssrc->duplicates;
 
 	// process per-second stats
 	int64_t cur_ts = rtpe_now;
 	int64_t last_sample;
-	int64_t sample_packets, sample_octets, sample_packets_lost, sample_duplicates;
+	int64_t sample_packets, sample_octets, sample_packets_lost, sample_packets_lost_rtcp, sample_duplicates;
 
 	// sample values
 	last_sample = atomic64_get_set(&ssrc->last_sample, cur_ts);
 	sample_packets = atomic64_get_set(&ssrc->sample_packets, packets);
 	sample_octets = atomic64_get_set(&ssrc->sample_octets, octets);
 	sample_packets_lost = atomic64_get_set(&ssrc->sample_packets_lost, packets_lost);
+	sample_packets_lost_rtcp = atomic64_get_set(&ssrc->sample_packets_lost_rtcp, packets_lost_rtcp);
 	sample_duplicates = atomic64_get_set(&ssrc->sample_duplicates, duplicates);
 
 	json_builder_set_member_name(json, "packets");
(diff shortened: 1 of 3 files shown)


rfuchs:
Commenting out that line would probably break reporting the loss based on just what was received via RTCP, no?

jmontorosf:
Hi Richard,

Thanks for you prompt reply, yeah it would break the individual ssrc rtcp-based metrics, but i just checked and the voip_metrics which are rtcp based as well seem to be working fine. the problem in individual ssrc is when rtpengine is measuring locally and and rtcp report comes in it reset the counters... so it produces a misleading value.... i think there should be either a way to select if i want locally measured metric or rtcp based metrics (currently i prefer locally, because is more granular and it's actually what rtpengine sees, additionally rtcp messages can get lost as well if the packet loss is high enough) or produce two metrics one for rtcp based to see what the other end reports, and another one to verify what rtpengine considers has been lost..

Does it make sense?

rfuchs:
Just to be clear, this is with the `measure-rtp` setting, right?

Have you tried combining this with local RTCP generation (`generate-rtcp` flag)? I'm a little hazy on the details, but IIRC enabling this switches RTCP processing to a different code path so that locally measured values take precedence? (But I could be wrong and maybe received RTCP is still taken into account.) IIRC there's also something about interval-reported stats that require RTCP generation (perhaps that's MOS only?) because only then the relevant timer is active.

The thing is that locally measured values can be different from values reported via RTCP, not only when the remote RTCP sender reports bogus values, but packets lost on the way from rtpengine to the receiver could only be known from RTCP reports and could not be measured locally. Both values are interesting and relevant, but can possibly mean different things.

No argument that things can be improved here, but just removing that line of code is not the solution.

jmontorosf:
I agree that removing that line is probably not the solution, it was more to ilustrate my findings...

i noticed that the --measure-rtp doesn't seem to have any effect in measuring packets loss... it produces locally measured values with or without it, with the generate-RTCP flag it somehow improves the data, it doesn't seem to reset the counters, but still produces artifacts for overcounting lost packets when there are incoming rtcp messages....

do you think could there be a way to split it in two different metrics? to avoid mixing them up or do think there could be another better solution??

Thanks a lot for your support

rfuchs:
> i noticed that the --measure-rtp doesn't seem to have any effect in measuring packets loss... it produces locally measured values with or without it

That depends on the call scenario and the config. With MQTT reporting, the option is always enabled implicitly, at least for the kernel module.

> with the generate-RTCP flag it somehow improves the data

The difference (probably?) being that the RTCP timer (`__rtcp_timer_run()`) runs and through `rtcp_send_report()` manages the internal report blocks.

> it doesn't seem to reset the counters, but still produces artifacts for overcounting lost packets when there are incoming rtcp messages....
> 
> do you think could there be a way to split it in two different metrics? to avoid mixing them up or do think there could be another better solution??

Easiest probably would be to just keep two separate values, one for values taken from RTCP reports and another for measured?

jmontorosf:
> Easiest probably would be to just keep two separate values, one for values taken from RTCP reports and another for measured?

agreed!

jmontorosf:
@rfuchs I pushed some changes, what are your thoughts on this?

rfuchs:
Looks reasonable

PR #2120, opened 2026-06-11
Backport "support 'all' codec accept mode"

orgads:
Backport to mr26.0.


orgads:
@rfuchs I'm not sure about your backport policy, but I'd love to have this in mr26.0, so I can use binary packages instead of maintaining patches.

rfuchs:
Normally we only backport bug fixes. I'll see if this qualifies or could be an exception.

PR #2125, opened 2026-06-27
Unref ports after re-subscribe

alexbakker:
Previously, the ports used for a monologue subscription would leak upon re-subscribe. ``__assign_stream_fds`` cleared the stream fd queue without decrementing the ref count. This prevented ``monologue_destroy`` from releasing the ports on unsubscribe.


alexbakker:
Looks like the ``daemon-tests-alias`` test was relying on ports staying open after monologue destroy, but this only worked because of the ref count issue. Ports are now correctly released, so sends to those ports will fail. I've replaced those assertions with a new ``snd_no`` helper.

PR #2141, opened 2026-07-18
Treat missing a=fmtp as defaults for AMR

alexbakker:
Per RFC 4867 section 8.1, absent fmtp parameters imply defaults (bandwidth-efficient mode, no CRC, no robust-sorting, no interleaving). Previously, codec matching was rejected when either side had no ``a=fmtp`` attribute for the AMR payload type, causing AMR to be dropped as a "stray answer codec".

This patch changes this behavior to instead "parse" fmtp even when absent so that the format struct is populated with zero-initialized defaults and fmtp_parsed is set.

### DIFF
--- a/t/test-transcode.c
+++ b/t/test-transcode.c
@@ -119,6 +119,7 @@ static void __start(const char *file, int line) {
 }
 
 #define transcode(codec) t_queue_push_tail(&flags.codec_transcode, sdup(#codec))
+#define transcode_s(s) t_queue_push_tail(&flags.codec_transcode, sdup(s))
 #define c_accept(codec) t_queue_push_tail(&flags.codec_accept, sdup(#codec))
 #define c_consume(codec) t_queue_push_tail(&flags.codec_consume, sdup(#codec))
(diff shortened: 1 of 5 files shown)


t/test-transcode.c
```diff
@@ -621,6 +621,48 @@ int main(void) {
 			packet_seq(A, 96, AMR_WB_payload_noe, 0, 0, -1, ""); // nothing due to resampling/decoding buffer
 			packet_seq_nf(A, 96, AMR_WB_payload_noe, 320, 1, 0, PCMU_payload);
 			end();
+
+			// forward AMR-WB, answer without fmtp (missing = defaults = octet-align=0)
+			// offer has octet-align=0, answer omits fmtp -> should match
+			start();
+			sdp_pt(0, PCMU, 8000);
+			transcode(AMR-WB/16000/1///octet-align=0);
```
rfuchs: doesn't compile because the `//` is interpreted as a comment

alexbakker:
@rfuchs I must have forgotten to push the latest revision. Sorry about that. The tests should pass now.

rfuchs:
Much better, thank you

PR #2142, opened 2026-07-18
recording-daemon: lifecycle notification events

goolanceman:
## Summary

Extends `rtpengine-recording` notifications from the historical **finished-only** path to a configurable **call/stream lifecycle** event set, while remaining **backward compatible**.

### Events (`notify-events` CSV)

| Token | Event name |
|-------|------------|
| `opened` | `recording_file_opened` |
| `started` | `recording_started` |
| `finished` (**default**) | `recording_finished` |
| `discarded` | `recording_discarded` |
| `failed` | `recording_failed` |
| `call-started` | `call_recording_started` |
| `call-finished` | `call_recording_finished` |
| `call-discarded` | `call_recording_discarded` |
| `all` | all of the above |

Default remains `finished` only so existing deployments are unchanged.

### New options

- `notify-events` — CSV list (default `finished`)
- `notify-json` — POST JSON body via json-glib
- `notify-no-metadata` — omit call metadata from payloads
- `notify-command-format` — `legacy` | `extended` | `json-env`
- `notify-queue-limit` — bound non-terminal queue depth (default 1000; `0` = unlimited). Terminal events are always accepted.

### Design

- Keeps upstream **action-based** notify model (`notif_action_t` / `notif_req_t`) and `lib/http.c` helpers
- Pure event helpers in `notify_events.c` (shared with unit tests)
- Stream hooks in `output.c`; call hooks in `metafile.c` / `packet.c`
- Snapshots strings into `notif_req` before queueing (workers never touch live objects)
- Metrics logged on shutdown: enqueued / success / retry / giveup / dropped

### Tests

- Unit: `t/test-notify-events.c` — PASS
- E2E smoke: `t/notify_e2e/` temp Python HTTP receiver + metafile lifecycle — **E2E_NOTIFY_OK**
  - Built/ran in Docker
  - Received `call_recording_finished` with JSON body + lifecycle headers

### Docs

- `docs/rtpengine-recording.md`
- `etc/rtpengine-recording.conf`

### Example

```
notify-uri = https://example.com/rec/events
notify-events = opened,started,finished,discarded,failed,call-started,call-finished,call-discarded
notify-json = true
notify-concurrency = 5
notify-retries = 10
```

### Compatibility

- Default config behavior is identical to previous finished-only notifications
- Legacy `notify-command` argv (`path`, `db_id`) preserved when `notify-command-format=legacy`

---

I used an AI assistant to help with some of the repetitive work (e.g. scaffolding, draft text), but the architecture, behaviour, and all final code changes were reviewed, adapted, and validated by me. The goal is to speed up the mechanical parts while still having a human developer take responsibility for correctness, style, and alignment with  rtpengine 's conventions. Feedback is very welcome and I'll iterate on the implementation as needed.


recording-daemon/notify.c
```diff
@@ -91,119 +156,98 @@ static void failed_http(notif_req_t *req) {
...
 	}
-
-	// clean up
-
 	pthread_mutex_unlock(&timer_lock);
 	pthread_mutex_destroy(&timer_lock);
 	pthread_cond_destroy(&timer_cond);
-
 	return NULL;
 }
```
rfuchs, line 273: Why remove the debug log messages and the existing comments?

recording-daemon/notify.c
```diff
@@ -91,119 +156,98 @@ static void failed_http(notif_req_t *req) {
...
-		return 1;
-	if (a < b)
-		return -1;
-	if (a > b)
-		return 1;
+	if (a->retry_time < b->retry_time) return -1;
+	if (a->retry_time > b->retry_time) return 1;
+	if (a < b) return -1;
+	if (a > b) return 1;
 	return 0;
```
rfuchs, line 286: Don't lump in reformats in the same commit

rfuchs (commented):
This is all AI generated, isn't it

goolanceman:
> This is all AI generated, isn't it

Good eye 🙂
I did use an AI assistant to help draft parts of this (especially the docs/text), but I've reviewed and adjusted everything myself and verified the code and tests. I'm fully responsible for the changes here, so if anything looks off, please call it out and I'll fix it.

recording-daemon/notify.c
goolanceman: Good catch, that removal was unintentional side‑effect from the refactor.

I have now restored the original comments and debug logging around the HTTP and command notify paths and the retry timer.

The functional changes remainn, but the surrounding comments and logs should now match the original intent much more closely.

recording-daemon/notify.c
goolanceman: You are right, some of the changes in  notify.c  ended up being style‑only and made the diff noisier than necessary. 

I have reverted the pure reformatting (e.g. the one‑line  notify_req_cmp()  and similar) so that the remainng diff in this file is focused on the behavioural changes: lifecycle events, metrics, queue limits, JSON payloads, and the S3/GCS/DB interaction.

recording-daemon/notify.c
```diff
@@ -305,27 +511,333 @@ static const notif_action_t command_action = {
...
+	req->event = event;
+	req->terminal = notify_event_is_terminal(event) ? 1 : 0;
+
+	if (mf) {
+		req->call_id = g_strdup(mf->call_id);
+		req->call_start = (double) mf->start_time_us / 1000000.;
+		req->call_db_id = mf->db_id;
+		if (!notify_no_metadata && mf->metadata)
+			req->metadata = g_strdup(mf->metadata);
+	}
```
rfuchs, line 680: This whole function seems redundant? It copes all the fields into a `notif_req_t`, only to have the fields copied again in `notify_enqueue()` and then freed?

recording-daemon/notify.c
```diff
@@ -213,64 +301,139 @@ void notify_setup(void) {
...
+	const char *full_filename = req->full_filename ? req->full_filename :
+		(o && o->full_filename ? o->full_filename : NULL);
+	const char *output_id = req->output_id ? req->output_id : (o && o->output_id ? o->output_id : NULL);
+	const char *metadata = req->metadata ? req->metadata :
+		(!notify_no_metadata && mf && mf->metadata ? mf->metadata : NULL);
+	double call_start = req->call_start > 0 ? req->call_start :
+		(mf ? (double) mf->start_time_us / 1000000. : 0);
+	double stream_start = req->stream_start > 0 ? req->stream_start :
+		(o ? (double) o->start_time_us / 1000000. : call_start);
+	unsigned long long call_db_id = req->call_db_id ? req->call_db_id : (mf ? mf->db_id : 0);
```
rfuchs, line 364: The call ID and the other fields are copied from the `metafile_t` or the output anyway. What's the reason for having all these case distinctions here?

recording-daemon/notify.c
```diff
@@ -213,64 +301,139 @@ void notify_setup(void) {
...
+				g_tree_remove(notify_timers, first);
+				nont_inflight_dec(first);
+				if (first->action && first->action->cleanup)
+					first->action->cleanup(first);
+				req_free_fields(first);
+				g_free(first->name);
+				g_free(first);
+			}
+			g_tree_destroy(notify_timers);
+			notify_timers = NULL;
```
rfuchs, line 329: This seems like a candidate for `g_tree_foreach` ?

rfuchs (commented):
Not sure if I'm willing to accept this as it all looks like an overly complicated wall of code, typical of AI slop

rfuchs:
Abandoned

Issue #2146, opened 2026-07-27
Crash on DTLS Shutdown


bundasmanu:
### rtpengine version the issue has been seen with

14.1.1.19+0~mr14.1.1.19

### Used distribution and its version

Ubuntu 24.04.4 LTS

### Linux kernel version used

6.17.0-1019-aws

### CPU architecture issue was seen on (see `uname -m`)

x86_64

### Expected behaviour you didn't see

When we do a normal call between two endpoints, on hangup, RTPEngine must not crash.

### Unexpected behaviour you saw

A crash on call hangup.

### Steps to reproduce the problem

WebRTC call.
- User A (WebRTC client) calls User B (that one is not a WebRTC Client);
- User B answer;
- Wait some time;
- User B hangups;

Right after hangup, the crash happens and RTPEngine restarts.
If A hangups the problem is not visible

### Additional program output to the terminal or logs illustrating the issue

```shell
`coredumpctl info rtpengine`

           PID: 1874 (rtpengine)
           UID: 111 (rtpengine)
           GID: 114 (rtpengine)
        Signal: 11 (SEGV)
     Timestamp: Fri 2026-07-24 21:57:01 UTC (35s ago)
  Command Line: /usr/bin/rtpengine -f -E --no-log-timestamps --pidfile /run/rtpengine/ngcp-rtpengine-daemon.pid --config-file /etc/rtpengine/rtpengine.conf
    Executable: /usr/bin/rtpengine
 Control Group: /system.slice/ngcp-rtpengine-daemon.service
          Unit: ngcp-rtpengine-daemon.service
         Slice: system.slice
       Boot ID: 58f5fd72364040119679619ae3edd1fa
    Machine ID: ec2b8be32c374ea7d41100467fb507ec
      Hostname: ip-10-34-148-105
       Storage: /var/lib/systemd/coredump/core.rtpengine.111.58f5fd72364040119679619ae3edd1fa.1874.1784930221000000.zst (present)
  Size on Disk: 10.7M
       Message: Process 1874 (rtpengine) of user 111 dumped core.

                Module libgomp.so.1 from deb gcc-14-14.2.0-4ubuntu2~24.04.1.amd64
                Module libgcc_s.so.1 from deb gcc-14-14.2.0-4ubuntu2~24.04.1.amd64
                Module libstdc++.so.6 from deb gcc-14-14.2.0-4ubuntu2~24.04.1.amd64
[...]

PR #2147, opened 2026-07-27
fix: add guard to avoid packet sent on shutdown on already dead streams

bundasmanu:
This PR is related to the issue: [2146](https://github.com/sipwise/rtpengine/issues/2146).

The fix results from gdb analysis, and uses a similar approach that was used previously on similar problems like: (https://github.com/linuxmaniac/rtpengine/commit/ceb7996ccadd03797ce4dd6de64dffe8c7c5caa1).

I already have tested it, and crashes are not happening anymore.

### DIFF
--- a/daemon/dtls.c
+++ b/daemon/dtls.c
@@ -646,6 +646,8 @@ static long dtls_bio_callback(BIO *bio, int oper, const char *argp, size_t len, 
 	struct stream_fd *sfd = d->sfd;
 	if (!sfd)
 		return ret;
+	if (!sfd->socket.family || sfd->socket.fd < 0)
+		return ret;
 
 	__DBG("dtls packet output: len %zu %02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x%02x",
 		len,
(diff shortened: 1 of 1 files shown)


daemon/dtls.c
```diff
@@ -646,6 +646,8 @@ static long dtls_bio_callback(BIO *bio, int oper, const char *argp, size_t len,
 	struct stream_fd *sfd = d->sfd;
 	if (!sfd)
 		return ret;
+	if (!sfd->socket.family || sfd->socket.fd < 0)
```
rfuchs: Suggest to change the test for open socket to `!= -1`

Technically I think the test on `fd` can be removed altogether, as (in theory) we guarantee the socket to be open if `family` is set

bundasmanu: So, if i understood, you mean to do a simple check like this? `if (!sfd->socket.family)`

rfuchs: That should be enough, yes

bundasmanu: I will test, and after i will you give feedback. Thank you.

**EDIT**: Worked, thank you.

PR #2167, opened 2026-09-04
Add `force strip extmap` flag

dnygate:
RTP header extensions are only processed, and unknown ones removed, if at
least one side of the call negotiated header extensions through `a=extmap`
attributes in the SDP. Otherwise RTP is forwarded untouched, including any
header extensions an endpoint sends without having announced them.

This adds a `force strip extmap` flag that turns on header extension
processing for all streams of a call regardless of whether any extensions
were negotiated, so that unannounced header extensions are removed from
forwarded RTP. The flag is remembered for the lifetime of the call, so it
only needs to be given once.

Use case: media coming from a WebRTC selective forwarding unit carries
header extensions that WebRTC clients use among themselves (audio level,
transport-wide congestion control) but which were never part of the SIP
side's SDP. Some SIP endpoints do not cope with unexpected header
extensions, and with this flag rtpengine can strip them selectively on
such routes.

Tests: a case in `t/auto-daemon-tests-rtp-ext.pl` sends RTP with header
extensions in both directions on a call without any `a=extmap` and checks
that they arrive without the extensions.

### DIFF
--- a/daemon/call.c
+++ b/daemon/call.c
@@ -2762,6 +2762,8 @@ static void __call_monologue_init_from_flags(struct call_monologue *ml, struct c
 
 	call->last_signal_us = rtpe_now;
 	call->deleted_us = 0;
+	if (flags->force_strip_extmap)
+		CALL_SET(call, FORCE_STRIP_EXTMAP);
 	call->media_rec_slots = (flags->media_rec_slots > 0 && call->media_rec_slots == 0)
 								? flags->media_rec_slots
 								: call->media_rec_slots;
--- a/daemon/call_flags.c
+++ b/daemon/call_flags.c
@@ -970,6 +970,10 @@ const char *call_ng_flags_flags(str *s, unsigned int idx, helper_arg arg) {
 		case CSH_LOOKUP("strip-extmap"):
 		case CSH_LOOKUP("strip extmap"):
 			return call_ng_flags_str_ht(STR_PTR("all"), 0, &out->rtpext_strip);
+		case CSH_LOOKUP("force-strip-extmap"):
+		case CSH_LOOKUP("force strip extmap"):
+			out->force_strip_extmap = true;
+			break;
 		case CSH_LOOKUP("symmetric-codecs"):
 		case CSH_LOOKUP("symmetric codecs"):
 			ilog(LOG_INFO, "Ignoring obsolete flag `symmetric-codecs`");
--- a/daemon/media_socket.c
+++ b/daemon/media_socket.c
@@ -2545,7 +2545,10 @@ static void __determine_rtpext_handler(struct call_media *in, struct call_media 
 	if (!sh || !out)
 		return;
 
-	if (in->extmap.length || out->extmap.length)
+	/* With `force strip extmap` set on the call, use the extmap printer even
+	 * if no header extensions were negotiated, so that unannounced header
+	 * extensions are removed from forwarded RTP. */
+	if (in->extmap.length || out->extmap.length || CALL_ISSET(out->call, FORCE_STRIP_EXTMAP))
 		sh->rtpext = &rtpext_printer_extmap;
 	else
 		sh->rtpext = &rtpext_printer_copy;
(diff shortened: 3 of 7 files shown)


rfuchs:
I feel like this should be handled unconditionally, without the need for an extra flag.

IOW if one side advertised some RTP extensions and the other side didn't accept them, then they should be removed anyway.

There could be reasons not to of course, but intuitively I would say that removing unsupported extensions should be the default behaviour, and possibly leaving them in place should be the optional alternative.

dnygate:
Understood. `__determine_rtpext_handler()` becomes an unconditional
`sh->rtpext = &rtpext_printer_extmap;`, and the flag itself, `CALL_FLAG_FORCE_STRIP_EXTMAP`,
the `__call_monologue_init_from_flags()` hook and the documentation section all come out with
it, which leaves the change as a net deletion. `rtpext_printer_copy` is still reached from
`media_player.c:814`, `sink_handler_set_generic()` and `media_socket.c:2618`, so it stays
where it is. I will rework the PR on that basis and drop the flag from the test.

On the inverse option, leaving unnegotiated extensions in place, I would rather not add it
until there is a concrete need for it, since an opt-out that nothing exercises tends to rot.
Easy enough to add later, or now if you would prefer it in from the start.

dnygate:
Before reworking this I built the unconditional version and ran it against the existing
suite. It breaks six scenarios in `t/auto-daemon-tests-rtp-ext.pl`, twelve assertions in
total: `unsolicited exts` and `unsolicited exts w tc`, each in its plain, SRTP and
DTLS variant. Every one of them sets up a call with no `a=extmap` on either side, sends RTP
carrying header extensions, and then asserts that those extensions arrive intact, which is
exactly the behaviour the flag was there to opt out of. They arrived with 6c5190a7 last July.
Nothing else moves: `daemon-tests-main`, `daemon-tests-bundle`, `daemon-tests-dtls`,
`daemon-tests-rtcp` and `daemon-tests-transform` all still pass, so the blast radius is
confined to those six.

That makes the change you are describing a deliberate reversal of what those tests pin
rather than a gap in the coverage, and inverting tests you wrote is your call rather than
mine. Two ways forward, depending on which you had in mind:

1. Unconditional stripping, with the six scenarios inverted to expect the extensions gone.
2. Unconditional stripping plus an opt-out flag, with the six scenarios given that flag so
   they carry on asserting pass-through.

Keeping the opt-in flag as originally submitted is still on the table as well, if you would
rather leave the default as it is.

My own preference is the second if the pass-through was intentional for endpoints that
depend on receiving what they were sent, and the first if those tests were only ever
recording what the code happened to do at the time. Say which one you want and I will push
it.

rfuchs:
Unsolicited header extensions should remain intact because RFC 8285 style extensions (`extmap`) are not the only game in town, and the header extension might carry other information and in a different format. This would apply to scenarios in which neither side advertised support for `extmap`,

If one side advertised `extmap` but the other one didn't, then extensions can be stripped.

dnygate:
The current default already does that. `__determine_rtpext_handler()` picks the extmap
printer whenever either side has extmap, so one side advertising and the other not already
strips, and with neither side advertising the copy printer leaves the block alone. Nothing
needs changing there.

The RFC 8285 point is the part I'd missed. With no `a=extmap` anywhere there's no reason to
assume the block is 8285 at all, so rewriting it could corrupt a profile-specific extension
that was perfectly valid on the wire. That's fair...

So this goes back to where it started: an opt-in flag for the case where neither side
advertised extmap but the extensions still have to go. That's the SFU route in the
description, where the extensions are known to be 8285 and the SIP endpoint can't cope with
them. I'll put your reasoning in the docs so it's clear why this is opt-in rather than the
default, and leave the six `unsolicited exts` scenarios alone.

Otherwise it's unchanged from what you reviewed, though I'm happy to rebase it onto current
master if you'd like it "freshened up". :)

rfuchs:
Ah, is that the use case you have? Header extensions are sent but `a=extmap` was never advertised? It's hard to read through all this LLM generated text and actually figure out what's it all about...

dnygate:
Yes, sorry, thought it helped speed up - but honestly just ends up "waffling"

That's the use case exactly. The SFU sends header extensions without ever advertising
`a=extmap`, and the SIP side doesn't either, so nothing gets stripped and the SIP endpoint
chokes on them.

rfuchs:
Ok, in that case an extra flag to enable this is indeed appropriate.

dnygate:
Perfect. Nothing to change on the branch then :)

rfuchs:
Merged as fdbc23c87 and 8046d0a24

PR #2169, opened 2026-09-07
Fix build with libavcodec older than 61.19

dnygate:
Noticed whilst I was here :D on libavcodec older than 61.19.

    lib/avc.c:73:39: error: ‘struct <anonymous>’ has no member named ‘codec’

The decoder struct has no `codec` member. The line above uses the `codec` argument, and so
should this one.


rfuchs:
Whoops, thanks

